PR #2, opened 2018-10-15
PMM-2990: Drop singletons generated by cobra.

arvenil:
* Easier to write parallel tests.
* A bit better separation of app from cobra commands.
* Tests.
* Docs and cleanups.


AlekSi:
Kamil, please stop working on pmm-agent. What you are doing is a waste of time and efforts. We will not have that CLI _at all_. Just – stop.

PR #16, opened 2019-03-07
PMM-3478: Update percona/pmm

DexterHD:


DexterHD:
@AlekSi One more update of `percona/pmm`, as I merged percona/pmm#286 also

PR #20, opened 2019-03-22
PMM-3594 Send real data to pmm-managed.

AlekSi:


AlekSi:
@DexterHD Please focus on Go code and overall architecture, and tests.
@askomorokhov Please focus on numbers correctness and tests.

Makefile
```diff
@@ -32,6 +32,11 @@ init:                           ## Installs tools to $GOPATH/bin (which is expec
 gen:                            ## Generate files.
 	go generate ./...
 
+gen-init:
```
BupycHuk: what does this command do?

agents/builtin/mysql/mysql.go
```diff
@@ -93,78 +114,179 @@ func (m *MySQL) Run(ctx context.Context) {
...
+		}
+
+		buckets[i] = b
 	}
 
-	for _, str := range structs {
-		ess := str.(*eventsStatementsSummaryByDigest)
+	startS := uint32(periodStart.Unix())
+	lengthS := uint32(periodLength.Seconds())
+	for i, b := range buckets {
```
BupycHuk: why do we use two loops instead of one?

agents/builtin/mysql/mysql.go
```diff
@@ -93,78 +114,179 @@ func (m *MySQL) Run(ctx context.Context) {
...
-		if ess.Digest == nil || ess.DigestText == nil {
-			m.l.Debugf("Skipping %s.", ess)
+	for digest, currentESS := range current {
+		prevESS := prev[digest]
+		if prevESS == nil {
+			prevESS = new(eventsStatementsSummaryByDigest)
+		}
+		count := float32(currentESS.CountStar - prevESS.CountStar)
+		switch {
+		case count == 0:
```
BupycHuk: is it possible to have the same CountStar value in prevESS and currentESS after Truncate?

Makefile
AlekSi: It generates initial model files (like models.go) that should be edited by hands.
We use only a few tables from performance schema, but it is likely that we will use more in the near future, that's why I kept it for now.

agents/builtin/mysql/mysql.go
AlekSi: Good point, that's a missed refactoring opportunity after functions were merged.

agents/builtin/mysql/mysql.go
AlekSi: Yes, but we can't differentiate those two situations.

Makefile
BupycHuk: Could you add a comment to makefile? Like `## Generate initial model files`

AlekSi: That comment will make it a public target – visible in `make` output. That target is intended to be private.

agents/builtin/mysql/mysql.go
```diff
@@ -93,78 +114,179 @@ func (m *MySQL) Run(ctx context.Context) {
...
+			{float32(currentESS.SumSortMergePasses - prevESS.SumSortMergePasses), &mb.MMergePassesSum, &mb.MMergePassesCnt},
+			{float32(currentESS.SumSortRange - prevESS.SumSortRange), &mb.MSortRangeSum, &mb.MSortRangeCnt},
+			{float32(currentESS.SumSortRows - prevESS.SumSortRows), &mb.MSortRowsSum, &mb.MSortRowsCnt},
+			{float32(currentESS.SumSortScan - prevESS.SumSortScan), &mb.MSortScanSum, &mb.MSortScanCnt},
+
+			{float32(currentESS.SumNoIndexUsed - prevESS.SumNoIndexUsed), &mb.MNoIndexUsedSum, &mb.MNoIndexUsedCnt},
+			{float32(currentESS.SumNoGoodIndexUsed - prevESS.SumNoGoodIndexUsed), &mb.MNoGoodIndexUsedSum, &mb.MNoGoodIndexUsedCnt},
+		} {
+			if p.value != 0 {
+				*p.sum = p.value
```
askomorokhov: IMHO, `p` should have more descriptive name here, to easier catch it purposes.

DexterHD:
@AlekSi Is it all about QAN? I think it is.
However, there are only 4 matches of this word and all these words are mentioned only in comments. 
It's weird for me as nothing says what this code is all about (I mean `agents/builtin/mysql`)? Maybe we should use more clean naming? IMHO packages and other stuff names should represent a behavior, not implementation. Maybe it should be `QAN` in package name or in `ServiceName`, I don't know.

agents/builtin/mysql/mysql.go
```diff
@@ -93,78 +125,184 @@ func (m *MySQL) Run(ctx context.Context) {
...
+			// convert picoseconds to seconds
+			{float32(currentESS.SumTimerWait-prevESS.SumTimerWait) / 1e12, &mb.MQueryTimeSum, &mb.MQueryTimeCnt},
+			{float32(currentESS.SumLockTime-prevESS.SumLockTime) / 1e12, &mb.MLockTimeSum, &mb.MLockTimeCnt},
+
+			{float32(currentESS.SumRowsAffected - prevESS.SumRowsAffected), &mb.MRowsAffectedSum, &mb.MRowsAffectedCnt},
+			{float32(currentESS.SumRowsSent - prevESS.SumRowsSent), &mb.MRowsSentSum, &mb.MRowsSentCnt},
+			{float32(currentESS.SumRowsExamined - prevESS.SumRowsExamined), &mb.MRowsExaminedSum, &mb.MRowsExaminedCnt},
+
+			{float32(currentESS.SumCreatedTmpDiskTables - prevESS.SumCreatedTmpDiskTables), &mb.MTmpDiskTablesSum, &mb.MTmpDiskTablesCnt},
+			{float32(currentESS.SumCreatedTmpTables - prevESS.SumCreatedTmpTables), &mb.MQueryTimeSum, &mb.MTmpTableCnt},
```
askomorokhov: I think here should be:
{float32(currentESS.SumCreatedTmpTables - prevESS.SumCreatedTmpTables), **&mb.MTmpTablesSum,** **&mb.MTmpTablesCnt**},

agents/builtin/mysql/mysql.go
```diff
@@ -49,42 +59,64 @@ type Params struct {
 // Change represents Agent status change _or_ QAN collect request.
 type Change struct {
```
DexterHD: I don't know, but I think it's bad when one thing in public API represents more than one meaning. These things are different by nature, isn't it? I mean it has different areas of responsibility. `AgentStatus` is about the agent lifecycle but `CollectRequest` is about collecting data.

I mean It's like some function with the name `AddOrJoin(a, b int) int`.

AlekSi:
> However, there are only 4 matches of this word and all these words are mentioned only in comments.

My editors shows 14 occurrences of "qan" in mysql.go 😄. But yeah, I think we should revisit names when we start adding slow log processing.

agents/builtin/mysql/mysql.go
AlekSi: Good catch! Fixed.

agents/builtin/mysql/mysql.go
AlekSi: `p` is a local loop variable, it has to be short. And I can't think of a better name.

agents/builtin/mysql/mysql.go
AlekSi: Indeed, there were two channels. But they were harder to use and much easier to misuse. Supervisor reads changes channel until it is closed. If there are two of them, and one is closed – should we wait for the second to close, or not? Should we drain the data channel before removing that Agent? etc etc. Supervisor code become much more complex.
Again, we could revisit it once we have more other Agents.

PR #21, opened 2019-03-25
PMM-3191 Local agent API.

DexterHD:


main.go
```diff
@@ -172,6 +186,116 @@ func workLoop(ctx context.Context, cfg *config.Config, l *logrus.Entry, client a
...
+	</body>
+	</html>
+	`)).Execute(&buf, handlers)
+	if err != nil {
+		l.Panic(err)
+	}
+
+	http.Handle("/", proxyMux)
+	http.HandleFunc("/debug", func(rw http.ResponseWriter, req *http.Request) {
+		rw.Write(buf.Bytes())
```
golangcibot: Error return value of `rw.Write` is not checked (from `errcheck`)

agentlocal/agent_local.go
```diff
@@ -18,15 +18,69 @@ package agentlocal
 
 import (
 	"context"
+	"sync"
 
+	"github.com/percona/pmm-agent/config"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

agentlocal/agent_local.go
```diff
@@ -18,15 +18,66 @@ package agentlocal
...
+	"github.com/percona/pmm/api/agentpb"
+
+	"github.com/percona/pmm-agent/config"
 )
 
+// AgentLocalServer represents local agent api server.
 type AgentLocalServer struct {
+	cfg *config.Config
+
+	mx             sync.RWMutex
```
AlekSi: Typically it is `rw`.

agentlocal/agent_local.go
```diff
@@ -18,15 +18,66 @@ package agentlocal
...
+
+// NewAgentLocalServer creates new local agent api server instance.
+func NewAgentLocalServer(cfg *config.Config) *AgentLocalServer {
+	return &AgentLocalServer{cfg: cfg}
+}
+
+// SetMetadata sets new values of ServerMetadata.
+func (als *AgentLocalServer) SetMetadata(md *agentpb.AgentServerMetadata) {
+	als.mx.RLock()
+	defer als.mx.RUnlock()
```
AlekSi: It should be `Lock()` and `Unlock()`. We set value there, not read it.

agentlocal/agent_local.go
```diff
@@ -18,15 +18,66 @@ package agentlocal
...
+	als.serverMetadata = md
 }
 
+func (als *AgentLocalServer) getMetadata() agentpb.AgentServerMetadata {
+	als.mx.Lock()
+	defer als.mx.Unlock()
```
AlekSi: It should be `RLock()` and `RUnlock()`. We read value there, not set it.

config/config.go
```diff
@@ -76,6 +79,8 @@ func application(cfg *Config) (*kingpin.Application, *string) {
 
 	app.Flag("id", "ID of this pmm-agent. [PMM_AGENT_ID]").
 		Envar("PMM_AGENT_ID").PlaceHolder("</agent_id/...>").StringVar(&cfg.ID)
+	app.Flag("port", "Agent local API port. [PMM_AGENT_PORT]").
```
AlekSi: Please rename it to `listen-port` or something else. We have another port next line.

config/config.go
```diff
@@ -53,8 +53,11 @@ type Ports struct {
 // Config represents pmm-agent's static configuration.
 //nolint:maligned
 type Config struct {
-	ID      string `yaml:"id"`
-	Address string `yaml:"address"`
+	ID       string `yaml:"id"`
+	Address  string `yaml:"address"`
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+	Port     uint16 `yaml:"port"`
```
AlekSi: Please rename it to `ListenPort` or something else. We have another port as a part of `Address`.
Maybe rename `Address` to `ServerAddress`.

main.go
```diff
@@ -258,5 +387,23 @@ func main() {
...
+
+	wg.Add(1)
+	go func() {
+		wg.Done()
+		runJSONServer(ctx, cfg, grpcAddr)
+	}()
+
+	wg.Wait()
+
+	workLoop(ctx, cfg, l, client, localServer)
```
AlekSi: Wait. How that can work? We start `workLoop` only after gRPC Server is done?
Does it actually work?

DexterHD: Yes, it works. I always run the code and test it. Look at the code above. `wg.Done()` called after goroutine is run not after gRPC server is done.

AlekSi: That is super weird and unexpected. What is the point of doing that? Why not just `go runGRPCServer(ctx, cfg, localServer, grpcAddr)`? `WaitGroup` is used for waiting and synchronisation, and there is no waiting of synchronisation.

PR #26, opened 2019-04-09
PMM-3758 Local API agents list.

DexterHD:


agents/supervisor/supervisor.go
```diff
@@ -264,10 +295,36 @@ func filter(existing, new map[string]agentpb.AgentParams) (toStart, toRestart, t
...
+			s.agentProcesses[newState.AgentId].lastChangedState = newState
+		}
+	case typeBuiltIn:
+		if _, ok := s.builtinAgents[newState.AgentId]; ok {
+			s.builtinAgents[newState.AgentId].lastChangedState = newState
+		}
+	}
+}
+
+func (s *Supervisor) changeBuildInLastState(newState agentpb.StateChangedRequest) {
```
golangcibot: U1000: func `(*Supervisor).changeBuildInLastState` is unused (from `unused`)

agentlocal/agent_local.go
```diff
@@ -53,6 +65,17 @@ func (als *AgentLocalServer) getMetadata() agentpb.AgentServerMetadata {
 	return *als.serverMetadata
 }
 
+func (als *AgentLocalServer) getAgentsList() (res []*agentlocalpb.AgentInfo, err error) {
+	als.rw.RLock()
+	defer als.rw.RUnlock()
+
+	if als.ag == nil {
+		panic("agentsGetter is nil. Use SetAgentsGetter(ag agentsGetter) to set it.")
```
AlekSi: Can we pass that parameter to constructor? In general, half-constructed objects are bad.

agentlocal/agent_local.go
```diff
@@ -53,6 +65,17 @@ func (als *AgentLocalServer) getMetadata() agentpb.AgentServerMetadata {
 	return *als.serverMetadata
 }
 
+func (als *AgentLocalServer) getAgentsList() (res []*agentlocalpb.AgentInfo, err error) {
```
AlekSi: In general, avoid named results, they have too many problems; for an extreme example: https://twitter.com/mvdan_/status/1115300377434443776
Let's say we use named results only when something fancy is done with them. For example, setting `err` with `defer`/`recover()`. When nothing fancy happens, do not use them.

agents/supervisor/supervisor.go
```diff
@@ -52,24 +53,26 @@ type Supervisor struct {
 	qanRequests   chan agentpb.QANCollectRequest
 	l             *logrus.Entry
 
-	m              sync.Mutex
+	m              sync.RWMutex
```
AlekSi: `rw`

agents/supervisor/supervisor.go
```diff
@@ -264,10 +295,36 @@ func filter(existing, new map[string]agentpb.AgentParams) (toStart, toRestart, t
 
 //nolint:golint
 const (
-	type_TEST_SLEEP agentpb.Type = 998 // process
-	type_TEST_NOOP  agentpb.Type = 999 // built-in
+	typeProcess agentpb.Type = 998
+	typeBuiltIn agentpb.Type = 999
+
+	type_TEST_SLEEP = typeProcess
```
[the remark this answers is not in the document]
AlekSi: Please revert that change. I will disable HoundCI – it has no use for Go code.

agentlocal/agent_local.go
DexterHD: > Can we pass that parameter to the constructor?  In general, half-constructed objects are bad.
Maybe, I don't know, but in general, there is nothing wrong with it. It's simple https://en.wikipedia.org/wiki/Dependency_injection#Setter_injection

agents/supervisor/supervisor.go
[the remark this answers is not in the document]
DexterHD: https://github.com/percona/pmm-agent/pull/26/files#diff-0a814983b1a7e4cc3bde85d3833310f2R309

agents/supervisor/supervisor.go
```diff
@@ -230,6 +233,34 @@ func (s *Supervisor) QANRequests() <-chan agentpb.QANCollectRequest {
 	return s.qanRequests
 }
 
+// AgentsList returns info for all agents was runned by supervisor.
+func (s *Supervisor) AgentsList() ([]*agentlocalpb.AgentInfo, error) {
```
golangcibot: (*Supervisor).AgentsList - result 1 (error) is always nil (from `unparam`)

agents/supervisor/supervisor.go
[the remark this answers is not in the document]
AlekSi: So we use the same type and value for both "class"/"group" and specific type? Let's not do that.

agents/supervisor/supervisor.go
```diff
@@ -96,9 +99,6 @@ func NewSupervisor(ctx context.Context, paths *config.Paths, ports *config.Ports
 
 // SetState starts or updates all agents placed in args and stops all agents not placed in args, but already run.
 func (s *Supervisor) SetState(state *agentpb.SetStateRequest) {
-	s.m.Lock()
-	defer s.m.Unlock()
```
AlekSi: Why we drop that global lock, but replace it with individual locks in each method? It looks like we can loose consistency that way, and also introduce more complexity.

agents/supervisor/supervisor.go
```diff
@@ -52,9 +53,12 @@ type Supervisor struct {
 	qanRequests   chan agentpb.QANCollectRequest
 	l             *logrus.Entry
 
-	m              sync.Mutex
+	rw             sync.RWMutex
 	agentProcesses map[string]*agentProcessInfo
 	builtinAgents  map[string]*builtinAgentInfo
+
+	amx           sync.Mutex
```
AlekSi: It looks like if we use RWMutex, we can have parallel `/local/Status` calls which is a good thing.

agents/supervisor/supervisor.go
```diff
@@ -52,9 +53,12 @@ type Supervisor struct {
...
+	agentStatuses map[string]inventorypb.AgentStatus
```
AlekSi: ❗️The big problem is that we never delete elements from that map. We should do it.

agentlocal/agent_local.go
```diff
@@ -78,19 +92,16 @@ func (als *AgentLocalServer) Status(ctx context.Context, req *agentlocalpb.Statu
...
-	// TODO: Add real AgentsInfo
-	//agentsInfo := &agentlocalpb.AgentInfo{
-	//	AgentId:   "001",
-	//	AgentType: agentpb.Type_MYSQLD_EXPORTER,
-	//	Status:    inventory.AgentStatus_RUNNING,
-	//	Logs:      []string{},
-	//}
+	agentsInfo := als.getAgentsList()
+	if agentsInfo == nil {
+		agentsInfo = []*agentlocalpb.AgentInfo{}
```
AlekSi: You actually can't deferential between nil slice and non-nil empty slice at protobuf level. Returning `nil` is fine.

agents/supervisor/supervisor.go
```diff
@@ -462,10 +513,9 @@ func (s *Supervisor) processParams(agentID string, agentProcess *agentpb.SetStat
 	return &processParams, nil
 }
 
+// stopAll stops all agents.
+// SHOULD be guarded by Mutex (s.rw).
```
AlekSi:
```suggestion
// SHOULD be called with s.rw held for writing.
```

agents/supervisor/supervisor.go
```diff
@@ -109,6 +115,7 @@ func (s *Supervisor) SetState(state *agentpb.SetStateRequest) {
 }
 
 // setAgentProcesses starts/restarts/stops Agent processes.
+// SHOULD be guarded by Mutex (s.rw).
```
AlekSi:
```suggestion
// SHOULD be called with s.rw held for writing.
```

agents/supervisor/supervisor.go
```diff
@@ -171,6 +180,7 @@ func (s *Supervisor) setAgentProcesses(agentProcesses map[string]*agentpb.SetSta
 }
 
 // setBuiltinAgents starts/restarts/stops built-in Agents.
+// SHOULD be guarded by Mutex (s.rw).
```
AlekSi:
```suggestion
// SHOULD be called with s.rw held for writing.
```

agents/supervisor/supervisor.go
DexterHD: Fixed

agents/supervisor/supervisor.go
```diff
@@ -127,18 +134,21 @@ func (s *Supervisor) setAgentProcesses(agentProcesses map[string]*agentpb.SetSta
...
 		agent := s.agentProcesses[agentID]
+
 		agent.cancel()
 		<-agent.done // wait before releasing port
 
 		if err := s.portsRegistry.Release(agent.listenPort); err != nil {
 			s.l.Errorf("Failed to release port: %s.", err)
 		}
 		delete(s.agentProcesses, agentID)
+		delete(s.agentStatuses, agentID)
```
AlekSi: Sorry, but now we also should hold `s.arw`.
I wonder if we can trigger this race with a test.

agents/supervisor/supervisor.go
```diff
@@ -127,18 +134,24 @@ func (s *Supervisor) setAgentProcesses(agentProcesses map[string]*agentpb.SetSta
 	// stop first to avoid extra load
 	for _, agentID := range toStop {
 		agent := s.agentProcesses[agentID]
+
```
AlekSi: So you added an empty line there…

agents/supervisor/supervisor.go
```diff
@@ -191,6 +205,10 @@ func (s *Supervisor) setBuiltinAgents(builtinAgents map[string]*agentpb.SetState
 		agent.cancel()
```
AlekSi: … but not there. Why?

agents/supervisor/supervisor.go
```diff
@@ -268,6 +316,15 @@ const (
 	type_TEST_NOOP  agentpb.Type = 999 // built-in
 )
 
+func (s *Supervisor) changeLastState(agentID string, newState inventorypb.AgentStatus) {
+	s.arw.Lock()
+	defer s.arw.Unlock()
+	if s.agentStatuses == nil {
```
AlekSi: This is not possible. After `s.agentStatuses` is set to `nil` in `stopAll()` this `Supervisor` instance should not be (and can't be reused). Let's remove this defensive measure because it only make it look like _maybe somehow_ this is possible.

agents/supervisor/supervisor.go
```diff
@@ -288,11 +345,13 @@ func (s *Supervisor) startProcess(agentID string, agentProcess *agentpb.SetState
 	done := make(chan struct{})
 	go func() {
 		for status := range process.Changes() {
-			s.changes <- agentpb.StateChangedRequest{
+			newState := agentpb.StateChangedRequest{
 				AgentId:    agentID,
 				Status:     status,
 				ListenPort: uint32(port),
 			}
+			s.changes <- newState
```
AlekSi: No empty line.

agents/supervisor/supervisor.go
```diff
@@ -331,10 +390,13 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentpb.SetState
 		go func() {
 			for change := range m.Changes() {
 				if change.Status != inventorypb.AgentStatus_AGENT_STATUS_INVALID {
-					s.changes <- agentpb.StateChangedRequest{
+					newState := agentpb.StateChangedRequest{
 						AgentId: agentID,
 						Status:  change.Status,
 					}
+
```
AlekSi: Empty line. ¯\_(ツ)_/¯

agents/supervisor/supervisor.go
```diff
@@ -462,10 +524,9 @@ func (s *Supervisor) processParams(agentID string, agentProcess *agentpb.SetStat
 	return &processParams, nil
 }
 
+// stopAll stops all agents.
+// SHOULD be called with s.rw held for writing.
 func (s *Supervisor) stopAll() {
-	s.m.Lock()
-	defer s.m.Unlock()
```
AlekSi: We need a lock there – we change `agentProcesses` and `builtinAgents`.

agents/supervisor/supervisor.go
DexterHD: I have added this because of that: https://travis-ci.com/percona/pmm-agent/jobs/191821047#L594
So there it's not `maybe`. Tests fail without it.

AlekSi: That's a problem in the code that this `if` just hides.
`agentStatuses` should be set to `nil` only when all agents are fully stopped and all state changes are sent. There we assign it to `nil` only to allocate again.
Alternatively, we could not set `agentStatuses` to `nil` at all. It is done to other maps to prevent accidental `Supervisor` reuse. Since we can't reuse it already (other maps are assigned to `nil`), we can skip that assignment for `agentStatuses`.

agentlocal/agent_local.go
```diff
@@ -78,22 +86,20 @@ func (als *AgentLocalServer) Status(ctx context.Context, req *agentlocalpb.Statu
...
 	return &agentlocalpb.StatusResponse{
 		AgentId:      als.cfg.ID,
 		RunsOnNodeId: md.AgentRunsOnNodeID,
 		ServerInfo:   srvInfo,
-		AgentsInfo:   []*agentlocalpb.AgentInfo{}, // TODO: Add real AgentsInfo
+		AgentsInfo:   agentsInfo,
 	}, nil
 }
 
+func (als *AgentLocalServer) Reload(ctx context.Context, req *agentlocalpb.ReloadRequest) (*agentlocalpb.ReloadResponse, error) {
```
golangcibot: exported method `AgentLocalServer.Reload` should have comment or be unexported (from `golint`)

PR #34, opened 2019-04-19
PMM-3718 Setup command.

AlekSi:


config/config.go
```diff
@@ -38,8 +70,8 @@ type Paths struct {
 	TempDir          string `yaml:"tempdir"`
 }
 
-// Lookup replaces paths with absolute paths.
-func (p *Paths) Lookup() {
+// lookup replaces paths with absolute paths.
+func (p *Paths) lookup() {
```
golangcibot: U1000: func `(*Paths).lookup` is unused (from `unused`)

agentlocal/agent_local.go
```diff
@@ -57,6 +56,7 @@ type supervisor interface {
 // client is a subset of methods of client.Client used by this package.
 // We use it instead of real type for testing and to avoid dependency cycle.
 type client interface {
```
DexterHD: I think it should be divided into 2 separate interfaces. As those have different areas of responsibility. 
Also, comments like in interfaces `subset of methods of X` looks weird because interfaces are about abstraction and behavior not about "part of concrete type".

- First can be something like `serverMetadataProvider`. With one method: `GetAgentServerMetadata()`. 
- Second is https://godoc.org/github.com/prometheus/client_golang/prometheus#Collector. Field name for the second interface can be something like `clientCollector`.

```go
type serverMetadataProvider interface {
    GetAgentServerMetadata() *agentpb.AgentServerMetadata
}

func NewServer(..., mDataStore serverMetadataProvider, clientCollector prometheus.Collector, ...) *Server {
...
}
```

commands/clients.go
```diff
@@ -0,0 +1,128 @@
...
+	transport.Transport.(*http.Transport).TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}
+
+	agentlocalpb.Default.SetTransport(transport)
+}
+
+type statusResult struct {
+	ConfigFilePath string
+}
+
+// status returns locally running pmm-agent status.
```
DexterHD:
```suggestion
// localStatus returns locally running pmm-agent status.
```

agentlocal/agent_local.go
AlekSi: > Also, comments like in interfaces subset of methods of X looks weird because interfaces are about abstraction and behavior not about "part of concrete type".

But we don't need abstraction there. The comment explains why we use interface there – not because we need abstraction. That really a client as much as possible.

DexterHD: @AlekSi Sorry, but I disagree. Because the purpose of the client is to communicate with the server, not provide some metadata and/or insights. For example, there is can be some queue or another component which can provide that metadata. Is side functionality of the client to provide these things. 

Also, I didn't see that usage of interfaces anywhere (I mean `interface is a subset of... for...` it's a strange comment in principle). It's strange for me in general. Maybe you can give me some examples of that usage in some popular projects or in best practices? 
For me, the interface means abstraction or a contract. In any language. (For golang, for example here is an abstract: https://golang.org/doc/effective_go.html#interfaces).

In the end, I think, we need this here not because we need client here, but because we need something that provides `server metadata` and needs something else that provides `client insights` which we expose to local API. We don't care actually who provide that metadata and insights (It is client or provider or some shared library. Doesn't matter.).

AlekSi: We don't need interface or abstraction there. We would use specific type instead of interface if we could. We don't need anything that walks like a duck – we need our duck. Extra abstraction and flexibility only would make that code harder to understand.

DexterHD: @AlekSi This all about **Loose coupling** and good dependency names. In my opinion, `client` is a bad dependency name in this context. Give me the reason please, why we could use the client here? Why exactly the client? Give me examples, best practices, something I can learn and become better if I wrong because now I don't understand why it should be `client`.

Or I don't know. Nevermind. It has no reason to talk about all this stuff because it has no business value. You are the architect I'm the developer. I can sit and just write code like a code monkey. Only thing is I don't understand, where is the teamwork here and why we should do this code review at all, what the reason if only one person makes all decisions and knows how to do right things?

AlekSi: > Why exactly the client?

If we could, Instance of `client.Client` type. Not some abstraction.

> For example, there is can be some queue or another component which can provide that metadata.

Do not overcomplicate things for the sake of good practices. You argue about abstractions, loose coupling, and good practices. I argue that simple code, especially in a simple statically-typed language like Go, beats every other good practice.
You probably could start from https://speakerdeck.com/tastapod/why-every-element-of-solid-is-wrong.

> I can sit and just write code like a code monkey.

You can't, and we both know that. And I'm not asking for that. 😄

DexterHD: What the definition of `simple code` basically? It's subjective. How you realize which code is simple and which is not? What the metric of the simple code?

PR #41, opened 2019-04-25
PMM-3923 Add test stubs and more fixes.

DexterHD:
…pmm-agent/pull/37.


agents/builtin/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/percona/pmm-agent/agents/builtin/mongodb/internal/status"
 )
 
-const (
+var (
 	DefaultInterval  = time.Duration(60 * time.Second)
 	ReportChanBuffer = 1000
```
AlekSi: Let's make it unexpected const.

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,81 @@
...
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/mgo.v2/bson"
+
+	"github.com/percona/pmm-agent/agents/builtin/mongodb/internal/profiler/aggregator"
+	"github.com/percona/pmm-agent/agents/builtin/mongodb/internal/report"
+	"github.com/percona/pmm-agent/agents/builtin/mongodb/internal/test/profiling"
+)
+
+func TestProfiler(t *testing.T) {
+	aggregator.DefaultInterval = time.Duration(5 * time.Second)
```
AlekSi: Let's restore previous value in `defer`.

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,81 @@
...
+	url := "mongodb://pmm-agent:root-password@127.0.0.1:27017"
+	p := profiling.New(url)
+
+	logrus.SetLevel(logrus.DebugLevel)
```
AlekSi: Let's either comment this out, or restore previous value in `defer`.

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,81 @@
...
+	dialInfo, err := pmgo.ParseURL(url)
+	fatalOnErr(t, err)
```
AlekSi: Import `"github.com/stretchr/testify/require"`, use `require.NoError(t, err)`.

agents/builtin/mongodb/internal/profiler/profiler_test.go
AlekSi: Can we also reduce that value to 2s? 1s?

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,81 @@
...
+
+	assert.Equal(tw.t, "INSERT peoples", actual.Buckets[0].Fingerprint)
+	assert.Equal(tw.t, "test", actual.Buckets[0].DDatabase)
+	assert.Equal(tw.t, "peoples", actual.Buckets[0].DSchema)
+	assert.Equal(tw.t, "test-id", actual.Buckets[0].AgentId)
+	assert.Equal(tw.t, qanpb.MetricsSource_MONGODB_PROFILER, actual.Buckets[0].MetricsSource)
+	assert.Equal(tw.t, 1, actual.Buckets[0].NumQueries)
+	assert.Equal(tw.t, 60, actual.Buckets[0].MResponseLengthSum)
+	assert.Equal(tw.t, 60, actual.Buckets[0].MResponseLengthMin)
+	assert.Equal(tw.t, 60, actual.Buckets[0].MResponseLengthMax)
```
AlekSi: Can we compare whole structs?

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,81 @@
...
+	err = p.Disable("test")
+	fatalOnErr(t, err)
+}
+
+type testWriter struct {
+	t *testing.T
+}
+
+func (tw *testWriter) Write(actual *report.Report) error {
+	assert.NotNil(tw.t, actual)
```
AlekSi: That should be `require.NotNil`, otherwise `actual.Buckets` below will panic.
There are other similar places where `require` should used.

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,81 @@
...
+	assert.Equal(tw.t, 60, actual.Buckets[0].MResponseLengthMin)
+	assert.Equal(tw.t, 60, actual.Buckets[0].MResponseLengthMax)
+	return nil
+}
+
+func fatalOnErr(t *testing.T, err error) {
+	if err != nil {
+		t.Fatal(err)
+	}
+}
```
AlekSi: Remove it.

agents/builtin/mongodb/internal/profiler/sender/senter_test.go
```diff
@@ -15,3 +15,44 @@
...
+		EndTs:   time.Now().Add(time.Second * 10),
+		Buckets: []*qanpb.MetricsBucket{{Queryid: "test"}},
+	}
+
+	repChan := make(chan *report.Report)
+	tw := &testWriter{t: t, expectedReport: expected}
+	snd := New(repChan, tw, logrus.WithField("component", "test-sender"))
+	err := snd.Start()
+	if err != nil {
+		t.Fatal(err)
```
AlekSi: require

agents/builtin/mongodb/internal/test/profiling/profiling.go
```diff
@@ -0,0 +1,185 @@
...
+	})
+}
+
+func (p *Profiling) DropAll() error {
+	return p.RunDB(func(session pmgo.SessionManager, dbname string) error {
+		return p.Drop(dbname)
+	})
+}
+
+func (p *Profiling) ResetAll() error {
```
AlekSi: Let's unexport or remove all methods we don't use (all expect Start and Stop?).

agents/builtin/mongodb/internal/profiler/aggregator/aggregator.go
DexterHD: Don't understand. :/ What does it means `unexpected const`?

AlekSi: Broken autocomplete. un-exported

agents/builtin/mongodb/internal/test/profiling/profiling.go
DexterHD: There are no `Start/Stop` methods. It's Profiler helper to enable and/or disable MongoDB profiler in tests.

agents/builtin/mongodb/internal/profiler/aggregator/aggregator.go
DexterHD: Whatever, I don't understand what exactly the problem is? Is it about some IDE or editor problem? I don't see any problems here. `go test` works fine. It compiles and runs. What should I do here?

AlekSi: It is about the exported package API. We should export only what is used.

```go
const reportChanBuffer = 1000
```

agents/builtin/mongodb/internal/test/profiling/profiling.go
AlekSi: Is it used anywhere?

DexterHD: I don't understand why should we remove public APIs if we don't use them. It doesn't make any sense for me. Let fork all third-party libraries we use and remove all methods we don't use from it, why not? 

Why should I drop it? Tomorrow If I would like to add some tests I need to add it one more time? No thanks. I won't do it.

AlekSi: Then un-export them if you think they will be useful later.

DexterHD: Why? Let's fork all libs we are using and un-export all public methods we don't use. What is the problem with that?

AlekSi: There is a problem in your argument – reductio ad absurdum. The code in this repository (outside of `vendor/`) is not a third-party code. This is also not a library package. It should not expose more than necessary. Those methods are not used. I would argue we should remove them, but if you think they could be useful later – fine, let's keep them, but not export them.

And I don't think this discussion worth the time we spend on it. Please un-export those methods.

DexterHD: The problem is, your arguments it's only your wishes they have no sense or value in the context of this task or review. The reason you ask about it is only that you can, not because this code doesn't work or doing something wrong. 

Also, I think it worth the time as at this point we have only one code review criteria: "You should like this changes". 

I disagree with that code review approach so I will challenge it. 

I don't understand why you get to the bottom of every unnecessary detail when we have more big problems in design and architecture. For the real problems, you always say `We can fix it later...`. But for this stupid thing you always rejecting the review.

```
This is also not a library package. It should not expose more than necessary.
```
All packages expose some public API and the caller depends on a package, not vice versa.
Which will be public is the decision of package creator, not package user.

```
fine, let's keep them, but not export them.
```
It also doesn't make any sense for me to do some methods private only because the caller doesn't use it. Where is it came from? I never have heard about that approach. It's insane.

DexterHD: Also as long as we have something in a repository like this: https://github.com/percona/pmm-managed/blob/PMM-2.0/services/mysql/mysql.go IMHO you have no rights to speak about something we should remove or unexport.

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,74 @@
...
+
+	"github.com/percona/pmm-agent/agents/builtin/mongodb/internal/profiler/aggregator"
+	"github.com/percona/pmm-agent/agents/builtin/mongodb/internal/report"
+)
+
+func TestProfiler(t *testing.T) {
+	defer func() {
+		aggregator.DefaultInterval = time.Duration(time.Minute)
+	}()
+	aggregator.DefaultInterval = time.Duration(time.Second)
```
AlekSi: It is better to do something like:
```go
defaultInterval := aggregator.DefaultInterval
aggregator.DefaultInterval = time.Duration(time.Second)
defer func() { aggregator.DefaultInterval = defaultInterval }
```

That way we restore the previous value, not the value specified in the other package which should be synced on change.

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,74 @@
...
+	err = sess.DB("test").DropDatabase()
+	require.NoError(t, err)
+
+	ms := &testWriter{t: t}
+	prof := New(dialInfo, dialer, logrus.WithField("component", "profiler-test"), ms, "test-id")
+	err = prof.Start()
+	require.NoError(t, err)
+
+	err = sess.DB("test").C("peoples").Insert(bson.M{"name": "Anton"}, bson.M{"name": "Alexey"})
+	require.NoError(t, err)
```
AlekSi: That one better be be `assert`, otherwise we call `prof.Start()`, but do not call `prof.Stop()` if there is insertion error.

agents/builtin/mongodb/internal/profiler/profiler_test.go
```diff
@@ -15,3 +15,74 @@
...
+type testWriter struct {
+	t *testing.T
+}
+
+func (tw *testWriter) Write(actual *report.Report) error {
+	require.NotNil(tw.t, actual)
+	assert.Equal(tw.t, 1, len(actual.Buckets))
+
+	expected := &qanpb.MetricsBucket{
+		Fingerprint:        "INSERT peoples",
```
AlekSi: It really generates SQL-like fingerprints? wow

DexterHD: Yep. Also, I found some docs about this. https://confluence.percona.com/display/~kamil.dziedzic/MongoDB+Profiler+for+PMM+and+PT#MongoDBProfilerforPMMandPT-Fingerprint

agents/builtin/mongodb/internal/profiler/profiler_test.go
DexterHD: Fixed. In the old code, it was a part of the configuration in constructors and it was more clear. But then I have removed not used parameters and only 2 of them remained. So I move it to package level constant and then in package level variable to have a possibility to change in in tests.

PR #45, opened 2019-05-06
Add exhaustiveness checks on our "sum types".

AlekSi:
https://github.com/percona/pmm/pull/344


agentlocal/agent_local_test.go
```diff
@@ -0,0 +1,76 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package agentlocal
+
+import (
+	"context"
+	"testing"
+
+	"github.com/percona/pmm/api/agentlocalpb"
+	"github.com/percona/pmm/api/inventorypb"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

Makefile
```diff
@@ -54,14 +56,31 @@ test-race:                      ## Run tests with race detector.
 test-cover:                     ## Run tests and collect coverage information.
 	go test $(TEST_FLAGS) -coverprofile=cover.out -covermode=count ./...
 
-check-license:                  ## Check that all files have the same license header.
+check:                          ## Run required checkers and linters.
 	go run .github/check-license.go
+	go-sumtype ./vendor/... ./...
```
BupycHuk: why don't we run `golangci-lint` here?

AlekSi: https://golangci.com runs it on CI.
During development, the most typical use is `golangci-lint run` which is not much longer than `make check`. But another common use-case is `golangci-lint run --new-from-rev=PMM-2.0` to check only new/changed code. I kind of want to promote both without introducing three Makefile targets (the third one for Travis CI that does not run `golangci-lint`).

client/channel/channel.go
```diff
@@ -36,6 +36,22 @@ const (
 	prometheusSubsystem = "channel"
 )
 
+// Request represents an request from server.
+// It is similar to agentpb.ServerMessage except it can contain only requests,
+// and the payload is already unwrapped (XXX instead of ServerMessage_XXX).
+type Request struct {
```
BupycHuk: maybe it's better to rename these two types to `ServerRequest` and `AgentResponse`, because agent can send request to pmm-managed as well.

Makefile
AlekSi: Another option is
```
check-travis:
  go run .github/check-license.go
  go-sumtype ./vendor/... ./...

check: check-travis
  golangci-lint run
```

But that leaves out `golangci-lint run --new-from-rev=PMM-2.0`.

PR #47, opened 2019-05-06
PMM-3978 Actions.

DexterHD:
At least basic architecture and abstractions were laid, and pt-summary action is working. 
I think we can review and merge the current implementation and start to extend it for new actions.


actions/runner.go
```diff
@@ -0,0 +1,75 @@
...
+package actions
+
+import (
+	"context"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+const defaultTimeout = time.Duration(time.Second * 10)
```
golangcibot: unnecessary conversion (from `unconvert`)

actions/pt_summary.go
```diff
@@ -0,0 +1,59 @@
...
+	return p.id
+}
+
+func (p *ptSummary) Name() string {
+	return p.name
+}
+
+func (p *ptSummary) Run(ctx context.Context) ([]byte, error) {
+	executable := path.Join(p.customPath, p.name)
+	cmd := exec.CommandContext(ctx, executable, parseArguments(p.params)...)
```
golangcibot: G204: Subprocess launched with variable (from `gosec`)

actions/pt_summary.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+type ptSummary struct {
+	id         uuid.UUID
+	name       string
+	customPath string
+	params     map[string]string
+}
+
+func NewPtSummary(params map[string]string, customPath string) Action {
```
golangcibot: exported function `NewPtSummary` should have comment or be unexported (from `golint`)

actions/runner.go
```diff
@@ -0,0 +1,75 @@
...
+
+// ActionResult represents an action result.
+type ActionResult struct {
```
golangcibot: U1000: type `ActionResult` is unused (from `unused`)

actions/runner.go
```diff
@@ -0,0 +1,75 @@
...
+	ID             uuid.UUID
+	Name           string
+	Error          error
+	CombinedOutput []byte
+}
+
+// Runner represents action runner.
+// Action runner is component that can run an Actions.
+type Runner struct {
```
golangcibot: U1000: type `Runner` is unused (from `unused`)

actions/runner.go
```diff
@@ -0,0 +1,75 @@
...
+	out chan ActionResult
+}
+
+// NewRunner returns new runner.
+func NewRunner() *Runner {
```
golangcibot: U1000: func `NewRunner` is unused (from `unused`)

actions/runner.go
```diff
@@ -0,0 +1,75 @@
...
+func (r *Runner) Run(a Action) {
+	go run(a, r.out)
+}
+
+// ActionReady returns channel that you can use to read action results.
+func (r *Runner) ActionReady() chan ActionResult {
+	return r.out
+}
+
+func run(a Action, out chan ActionResult) {
```
golangcibot: U1000: func `run` is unused (from `unused`)

Gopkg.toml
```diff
@@ -16,7 +16,7 @@ required = [
 # TODO switch to tags before PMM 2.0 release
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "PMM-3978-actions"
```
AlekSi: Please create draft PR for that branch.

actions/actions.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"context"
+	"errors"
+
+	"github.com/google/uuid"
+)
+
+const (
+	ActionPtSummary = "pt-summary"
```
golangcibot: exported const ActionPtSummary should have comment (or a comment on this block) or be unexported (from `golint`)

client/channel/channel.go
```diff
@@ -39,7 +39,8 @@ const (
 // Channel encapsulates two-way communication channel between pmm-managed and pmm-agent.
 //
 // All exported methods are thread-safe.
-type Channel struct { //nolint:maligned
+type Channel struct {
```
golangcibot: struct of size 136 bytes could be of size 128 bytes (from `maligned`)

actions/shell_action.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"context"
+	"os/exec"
+	"sync"
+)
+
+type shellAction struct {
+	id         string
+	name       int32
```
golangcibot: `name` is unused (from `structcheck`)

actions/shell_action.go
```diff
@@ -0,0 +1,90 @@
...
+	command    string
+	customPath string
```
golangcibot: `customPath` is unused (from `structcheck`)

actions/concurrent_runner.go
```diff
@@ -0,0 +1,115 @@
...
+type ConcurrentRunner struct {
+	//nolint:unused
+	out    chan ActionResult
+	logger logrus.FieldLogger
+
+	rw      sync.RWMutex
+	actions map[string]Action
+}
+
+// NewRunner returns new runner.
```
golangcibot: comment on exported function `NewConcurrentRunner` should be of the form `NewConcurrentRunner ...` (from `golint`)

actions/concurrent_runner.go
```diff
@@ -0,0 +1,115 @@
...
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const defaultTimeout = time.Second * 10
+
+// ActionResult represents an action result.
+type ActionResult struct {
```
golangcibot: U1000: type `ActionResult` is unused (from `unused`)

actions/concurrent_runner.go
```diff
@@ -0,0 +1,115 @@
...
+	//nolint:unused
+	ID             string
+	Name           string
+	Error          error
+	CombinedOutput []byte
+}
+
+// ConcurrentRunner represents concurrent action runner.
+// Action runner is component that can run an Actions.
+type ConcurrentRunner struct {
```
golangcibot: U1000: type `ConcurrentRunner` is unused (from `unused`)

actions/concurrent_runner_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func(t *testing.T, ready <-chan ActionResult) {
+		defer wg.Done()
+		expected := []string{"signal: killed", "signal: killed"}
+		expectedOut := []string{"", ""}
+		for i := 0; i < 2; i++ {
+			a := <-ready
+			assert.Contains(t, expected, string(a.Error.Error()))
```
golangcibot: unnecessary conversion (from `unconvert`)

client/client.go
```diff
@@ -207,6 +214,35 @@ func (c *Client) processSupervisorRequests() {
...
 
+func (c *Client) sendActionResults() {
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		for ar := range c.concurrentActionRunner.ActionReady() {
+			var errCode int32 = 0
```
golangcibot: should drop = 0 from declaration of var errCode; it is the zero value (from `golint`)

actions/concurrent_runner.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (r *ConcurrentRunner) run(a Action) { //nolint:unused
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+	r.actions.Store(a.ID(), a)
+	actionFields := logrus.Fields{"id": a.ID(), "name": a.Name()}
+	r.logger.WithFields(actionFields).Debugf("Running action...")
+
+	select {
+	case <-ctx.Done():
```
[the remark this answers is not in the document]
DexterHD: I've tested it by hands and it works. But there is a problem, as when I cancel the task, there is nothing going to the channel. So I've changed this code a little bit, have made a test for this case, thanks.

client/client.go
```diff
@@ -220,6 +256,40 @@ func (c *Client) processChannelRequests() {
 			c.supervisor.SetState(p)
 			responsePayload = new(agentpb.SetStateResponse)
 
+		case *agentpb.StartActionRequest:
+			var a actions.Action
+			switch p.Name {
```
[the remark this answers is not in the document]
DexterHD: I think no. 

As you can see, it called `ConcurrentRunner` and his main task is to Run an Action concurrently, not creating them by name defined in API. A `Runner` gets an `Action` as a parameter, and it placed in the same package with actions. 

My idea here is that the `action` package is the `application layer`(inside ring), and the client is the `infrastructure layer`(outside ring). So the client can use the application layer code, but the application layer shouldn't depend on API (on `*agentpb.StartActionRequest` for example).

What do you think?

actions/concurrent_runner.go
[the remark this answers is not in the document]
BupycHuk: tbh, I think that the test doesn't test these lines. Have you run it with coverage?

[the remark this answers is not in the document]
DexterHD: Hmmm... You are right :) I will check it one more time. I think I'm really doing some strange here...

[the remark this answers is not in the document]
DexterHD: Yep. You were right. It's dead code. I've removed it and improve test a little bit. Now it works fine. It stops by timeout and removes actions from internal actions map properly.

client/client.go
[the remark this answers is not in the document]
BupycHuk: yeah, you are right that it shouldn't be part of the action runner, but maybe there should be another layer to be more consistent with the previous code.

actions/concurrent_runner_test.go
```diff
@@ -0,0 +1,80 @@
...
+		expected := []string{"test\n", "test2\n"}
+		for i := 0; i < 2; i++ {
+			a := <-ready
+			assert.Contains(t, expected, string(a.CombinedOutput))
+		}
+	}(t, cr.ActionReady())
+	wg.Wait()
+}
+
+func TestConcurrentRunnerTimeout(t *testing.T) {
```
BupycHuk: maybe we should add a test for Stop() method?

client/client.go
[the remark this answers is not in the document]
DexterHD: > but maybe there should be another layer to be more consistent with the previous code
You are right, I agree, it will be better to move action creation somewhere. 

I thought about that. The reason I didn't do this, is because we don't have at least several real actions to realize where we can move that code. Also, I thought about deadlines, so it was a balanced decision.

actions/shell_action.go
```diff
@@ -0,0 +1,100 @@
...
+// NewShellAction creates Shell action.
+//
+// Shell action, it's an abstract action that can run some "predefined" set of shell commands.
+// This commands can be a shell script, script written on interpreted language, or binary file.
+func NewShellAction(id string, command string, arg []string) Action {
+	return &shellAction{
+		id:      id,
+		command: command,
+		arg:     arg,
+		forbidden: map[string]struct{}{
```
rnovikovP: For security reasons, we can't have this in place. we need an only list(mapping) from command to bash script. No  direct name matching

DexterHD: @rnovikovP Yes but this binary path and name are configurable (They can be set with `--paths-...` flags), So we don't know how it will be called. 

Look how it used, here we already have that mapping you want:

https://github.com/percona/pmm-agent/pull/47/files#diff-1cd41a2611da49ce63630d5a51bd82c5R262

DexterHD: It seems this check is redundant at all as we check action type at a different level. Also, I use some commands like `sleep` and `echo` for testing. We can leave it as-is, remove it, or I can do something like this:

```golang
return &shellAction{
		id:      id,
		command: command,
		arg:     arg,
		allowed: map[string]struct{}{
			"pt-summary":       {},
			"pt-mysql-summary": {},
			"sleep":            {}, // for testing
			"echo":             {}, // for testing
		},
	}

	_, file := path.Split(p.command)
	if _, ok := p.allowed[file]; !ok {
		return nil, errUnknownAction
	}
```
But then we should change this internal library implementation all the time we add a new action, and I think it's a bad thing. Also, it breaks the possibility to give custom names to shell commands actions.

rnovikovP: 1. This type of verification is confusing. It  looks like  you  can pass anything and some developer later can expect this  
1. Negative verification if what we need to use only in rare cases. This is very dangerous
1. If we don't need it at all ...maybe better to remove 
1. If there were good to have some Action without parameters and not dependent on binary installed on the Client, i would suggest having something like `uname` command.  It will be useful later ;)

actions/actions.go
```diff
@@ -0,0 +1,40 @@
...
+
+package actions
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+)
+
+var errUnknownAction = errors.New("unknown action")
```
golangcibot: `errUnknownAction` is unused (from `varcheck`)

actions/shell_action_test.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func TestRunShellAction(t *testing.T) {
+	// setup
+	id := "/action_id/6a479303-5081-46d0-baa0-87d6248c987b"
+	cmd := "echo"
+	p := NewShellAction(id, cmd, nil)
+	_, err := exec.LookPath(cmd)
+	if err != nil {
+		t.Skipf("Test skipped, reason: %s", err)
```
AlekSi: We can rely on both `echo` and `sleep` being present. There is no need to lookup it and skip the test.

actions/actions.go
```diff
@@ -0,0 +1,36 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package actions
```
AlekSi: Please add package comment.

actions/actions.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+)
+
+// Action describe abstract thing that can be runned by a client and returns some output.
+// Every structure that implement this interface can be an action.
+type Action interface {
+	// ID returns an action UUID. Used in log messages.
+	ID() string
+	// String representation of action name. Used in log messages.
+	Name() string
```
AlekSi: Type.

actions/shell_action.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Run starts an action.
+// Action runs with internal CancelContext (by default), and can be stopped with Stop() method.
+// This method is blocking.
+func (p *shellAction) Run(ctx context.Context) ([]byte, error) {
+	p.mx.Lock()
+	p.ctx, p.cancel = context.WithCancel(ctx)
+	p.mx.Unlock()
+
+	cmd := exec.CommandContext(p.ctx, p.command, p.arg...) //nolint:gosec
```
AlekSi: See https://github.com/percona/pmm-agent/blob/master/agents/process/process_sys_linux.go
We should do the same thing there.

actions/concurrent_runner.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+const defaultTimeout = time.Second * 10
+
+// ActionResult represents an action result.
+//nolint:unused
+type ActionResult struct {
+	ID             string
+	Name           string
```
AlekSi: Type.

actions/concurrent_runner.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// ConcurrentRunner represents concurrent action runner.
+// Action runner is component that can run an Actions.
+//nolint:unused
+type ConcurrentRunner struct {
+	out    chan ActionResult
+	logger logrus.FieldLogger
+
+	actions sync.Map // map[string]Action
```
AlekSi: `sync.Map` for special cases only. It is not a generic-purpose map+Lock replacement. https://www.youtube.com/watch?v=C1EtfDnsdDs

actions/concurrent_runner.go
```diff
@@ -0,0 +1,100 @@
...
+	Error          error
+	CombinedOutput []byte
+}
+
+// ConcurrentRunner represents concurrent action runner.
+// Action runner is component that can run an Actions.
+//nolint:unused
+type ConcurrentRunner struct {
+	out    chan ActionResult
+	logger logrus.FieldLogger
```
AlekSi: We consistently use `l *logrus.Entry`.

actions/concurrent_runner.go
```diff
@@ -0,0 +1,100 @@
...
+		logger:  l,
+		timeout: timeout,
+		out:     make(chan ActionResult),
+	}
+}
+
+// Run runs an Action in separate goroutine.
+// When action is ready those output writes to ActionResult channel.
+// You can get all action results with ActionReady() method.
+func (r *ConcurrentRunner) Run(a Action) {
```
AlekSi: "Run" is the whole interval from start to finish. We do not wait for the action to finish, so we should call this method `Start`.

actions/concurrent_runner.go
AlekSi: Why not just `Runner`? We don't have a other runners.

actions/concurrent_runner_test.go
```diff
@@ -0,0 +1,80 @@
...
+	wg.Add(1)
+	go func(t *testing.T, ready <-chan ActionResult) {
+		defer wg.Done()
+		expected := []string{"test\n", "test2\n"}
+		for i := 0; i < 2; i++ {
+			a := <-ready
+			assert.Contains(t, expected, string(a.CombinedOutput))
+		}
+	}(t, cr.ActionReady())
+	wg.Wait()
```
AlekSi: There is no need for goroutine and WaitGroup.
```go
expected := []string{"test\n", "test2\n"}
for i := 0; i < 2; i++ {
	a := <-cr.ActionReady()
	assert.Contains(t, expected, string(a.CombinedOutput))
}
```

actions/concurrent_runner_test.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		// check action was deleted from actions map.
+		_, ok := cr.actions.Load(a1.ID())
+		_, ok2 := cr.actions.Load(a2.ID())
+		assert.False(t, ok)
+		assert.False(t, ok2)
+
+	}(t, cr.ActionReady())
+	wg.Wait()
```
AlekSi: Like above, goroutine and WaitGroup do nothing there.

client/client.go
```diff
@@ -75,11 +78,12 @@ type Client struct {
...
 func New(cfg *config.Config, supervisor supervisor) *Client {
 	return &Client{
-		cfg:        cfg,
-		supervisor: supervisor,
-		l:          logrus.WithField("component", "client"),
-		backoff:    backoff.New(backoffMinDelay, backoffMaxDelay),
-		done:       make(chan struct{}),
+		cfg:                    cfg,
+		supervisor:             supervisor,
+		concurrentActionRunner: actions.NewConcurrentRunner(logrus.WithField("component", "actions.Runner"), 0),
```
AlekSi: Just `"runner"`.

client/client.go
```diff
@@ -152,7 +156,7 @@ func (c *Client) Run(ctx context.Context) error {
 	//    It exits when an unexpected message is received from the channel, or when can't be received at all.
 	//    When Run is left, caller stops supervisor, and that allows processSupervisorRequests to exit.
 	// Done() channel is closed when both goroutines exited.
-	oneDone := make(chan struct{}, 2)
+	oneDone := make(chan struct{}, 3)
```
AlekSi: Please read the comment above, then update it and code accordingly.

client/client.go
```diff
@@ -207,6 +215,32 @@ func (c *Client) processSupervisorRequests() {
...
 
+func (c *Client) sendActionResults() {
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		for ar := range c.concurrentActionRunner.ActionReady() {
+			errMessage := ""
```
AlekSi: https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style
```go
var errMessage error
```

client/client.go
```diff
@@ -207,6 +215,32 @@ func (c *Client) processSupervisorRequests() {
...
+				ActionId: ar.ID,
+				Done:     ar.Error != nil,
+				Error:    errMessage,
+				Output:   ar.CombinedOutput,
+			})
+		}
+		c.l.Debugf("actionRunner ActionReady() channel drained.")
+	}()
+
+	wg.Wait()
```
AlekSi: Goroutine and WaitGroup do nothing there.

client/client.go
```diff
@@ -221,10 +255,34 @@ func (c *Client) processChannelRequests() {
 			responsePayload = new(agentpb.SetStateResponse)
 
 		case *agentpb.StartActionRequest:
-			panic("TODO")
+			var a actions.Action
+			switch p.Type {
+			case managementpb.ActionType_PT_SUMMARY:
+				processParams := p.Params.(*agentpb.StartActionRequest_ProcessParams_)
```
AlekSi: `GetProcessParams()`

client/client.go
```diff
@@ -221,10 +255,34 @@ func (c *Client) processChannelRequests() {
...
+				a = actions.NewShellAction(p.ActionId, c.cfg.Paths.PtSummaryAction, processParams.ProcessParams.Args)
+				c.concurrentActionRunner.Run(a)
+				responsePayload = &agentpb.StartActionResponse{}
```
AlekSi:
```go
responsePayload = new(agentpb.StartActionResponse)
```
Like above.

client/client.go
```diff
@@ -221,10 +255,34 @@ func (c *Client) processChannelRequests() {
...
+				// TODO: Implement explain action.
+				c.l.Errorf("not implemented action EXPLAIN")
+				continue
+
+			case managementpb.ActionType_ACTION_TYPE_INVALID:
+				c.l.Errorf("Unsupported action: %s.", p.Type)
+				continue
+			}
 
+		// Handle Action Cancel request from pmm-managed
```
AlekSi: All other cases in this switch are for requests from pmm-managed too. Why single out this one?

actions/concurrent_runner.go
```diff
@@ -0,0 +1,100 @@
...
+			r.actions.Delete(id)
+		}
+	}
+}
+
+func (r *ConcurrentRunner) run(a Action, t time.Duration) { //nolint:unused
+	ctx, cancel := context.WithTimeout(context.Background(), t)
+	defer cancel()
+	r.actions.Store(a.ID(), a)
+	actionFields := logrus.Fields{"id": a.ID(), "name": a.Name()}
```
AlekSi:
```go
l := r.l.WithFields(…)
```
Then use `l`.

client/client.go
```diff
@@ -207,6 +215,32 @@ func (c *Client) processSupervisorRequests() {
...
+
+		for ar := range c.concurrentActionRunner.ActionReady() {
+			errMessage := ""
+			if ar.Error != nil {
+				errMessage = ar.Error.Error()
+			}
+
+			c.channel.SendRequest(&agentpb.ActionResultRequest{
+				ActionId: ar.ID,
+				Done:     ar.Error != nil,
```
AlekSi: That's not correct: if action finished without error, `ar.Error` will be `nil`.

config/config_test.go
```diff
@@ -95,11 +95,13 @@ func TestGet(t *testing.T) {
...
-				NodeExporter:     "node_exporter",
-				MySQLdExporter:   "mysqld_exporter",
-				MongoDBExporter:  "mongodb_exporter",
-				PostgresExporter: "postgres_exporter",
-				TempDir:          os.TempDir(),
+				NodeExporter:         "node_exporter",
+				MySQLdExporter:       "mysqld_exporter",
+				MongoDBExporter:      "mongodb_exporter",
+				PostgresExporter:     "postgres_exporter",
+				PtSummaryAction:      "pt-summary",
```
golangcibot: unknown field `PtSummaryAction` in struct literal (from `typecheck`)

config/config_test.go
```diff
@@ -95,11 +95,13 @@ func TestGet(t *testing.T) {
...
+				PtMySQLSummaryAction: "pt-mysql-summary",
```
golangcibot: unknown field `PtMySQLSummaryAction` in struct literal (from `typecheck`)

actions/concurrent_runner.go
DexterHD: https://golang.org/pkg/sync/#Map
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or **(2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.** In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

It's **2** isn't it? It uses inside multiple goroutines, and it read/write disjoint sets of keys.

actions/shell_action.go
DexterHD: What is this, and how it works? Can you explain?

actions/actions.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Package actions provides PMM Actions implementations and related utils.
+package actions
+
+import (
+	"context"
+)
+
+// Action describe abstract thing that can be running by a client and returns some output.
+// Every structure that implement this interface can be an action.
```
AlekSi: Do we actually want this? Why? Why we export this interface at all?

actions/concurrent_runner.go
AlekSi: May you check if you added this type before it actually was used, and that `//nolint:unused` is no longer needed?

actions/concurrent_runner.go
DexterHD: Because such name better reflects the purpose.

client/client.go
```diff
@@ -144,15 +149,17 @@ func (c *Client) Run(ctx context.Context) error {
...
-	// We start two goroutines, and terminate the gRPC connection and exit Run when any of them exits:
+	// We start three goroutines, and terminate the gRPC connection and exit Run when any of them exits:
 	// 1. processSupervisorRequests reads requests (status changes and QAN data) from the supervisor and sends them to the channel.
 	//    It exits when the supervisor is stopped.
 	//    When the gRPC connection is terminated on exiting Run, processChannelRequests exits too.
 	// 2. processChannelRequests reads requests from the channel and processes them.
 	//    It exits when an unexpected message is received from the channel, or when can't be received at all.
 	//    When Run is left, caller stops supervisor, and that allows processSupervisorRequests to exit.
+	// 3. sendActionResults reads action results from action runner and sends them to pmm-managed.
+	//    It exits when messages can't be received at all or when appContext is done.
```
AlekSi: Note that other cases describe what happens next. That one should too.

client/client.go
```diff
@@ -207,6 +218,23 @@ func (c *Client) processSupervisorRequests() {
...
+func (c *Client) sendActionResults() {
+	for ar := range c.runner.ActionReady() {
+		var errMessage string
+		if ar.Error != nil {
+			errMessage = ar.Error.Error()
+		}
+
+		c.channel.SendRequest(&agentpb.ActionResultRequest{
+			ActionId: ar.ID,
+			Done:     ar.Error == nil,
```
AlekSi: Action is done when it is done, not when it is done without error (or with error as in the previous version of this line).

actions/concurrent_runner.go
```diff
@@ -0,0 +1,113 @@
...
+
+// ConcurrentRunner represents concurrent action runner.
+// Action runner is component that can run an Actions.
+//nolint:unused
+type ConcurrentRunner struct {
+	out chan ActionResult
+	l   *logrus.Entry
+
+	// runningActions stores CancelFunc's for running actions.
+	runningActions sync.Map // map[string]CancelFunc
```
AlekSi: When we are talking about a few actions, the fastest data structure is actually array/slice, not map or sync.Map. The latter is used when both the number of elements and number of CPU cores accessing them. That's hardly our case.
But we also don't care about performance in this case. What we care about is type safety. `interface{}` is not type safe.

client/client.go
```diff
@@ -221,10 +249,34 @@ func (c *Client) processChannelRequests() {
...
+				// TODO: Implement explain action.
+				c.l.Errorf("not implemented action EXPLAIN")
+				continue
+
+			case managementpb.ActionType_ACTION_TYPE_INVALID:
+				c.l.Errorf("Unsupported action: %s.", p.Type)
+				continue
+			}
 
+		// Handle Action Stop request.
```
AlekSi: Please remove that comment.

actions/shell_action.go
AlekSi: https://jira.percona.com/browse/PMM-3173

client/client.go
DexterHD: I don't understand what do you want. It's your code, not mine. Add code suggestion if you know how it should be.

client/client.go
```diff
@@ -144,15 +149,17 @@ func (c *Client) Run(ctx context.Context) error {
...
+	// We start three goroutines, and terminate the gRPC connection and exit Run when any of them exits:
 	// 1. processSupervisorRequests reads requests (status changes and QAN data) from the supervisor and sends them to the channel.
 	//    It exits when the supervisor is stopped.
 	//    When the gRPC connection is terminated on exiting Run, processChannelRequests exits too.
 	// 2. processChannelRequests reads requests from the channel and processes them.
 	//    It exits when an unexpected message is received from the channel, or when can't be received at all.
 	//    When Run is left, caller stops supervisor, and that allows processSupervisorRequests to exit.
+	// 3. sendActionResults reads action results from action runner and sends them to pmm-managed.
+	//    It exits when messages can't be received at all or when appContext is done.
 	// Done() channel is closed when both goroutines exited.
```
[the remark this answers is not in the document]
DexterHD: ???

actions/actions.go
DexterHD: ???

actions/concurrent_runner.go
DexterHD: I don't understand what are you talking about? About premature optimization or about what? If we are talking about speed, linear search in HashMap is `O(1)` and on the array is `O(N)`. What the problem? 

The latter is used in 2 cases:

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or **(2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.** In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

Why this code is not the 2nd case?

client/client.go
AlekSi: That's _our_ code.

client/client.go
```diff
@@ -144,25 +149,33 @@ func (c *Client) Run(ctx context.Context) error {
...
-	// 2. processChannelRequests reads requests from the channel and processes them.
+	// 2. sendActionResults reads action results from action runner and sends them to the channel.
+	//    It exits when the action runner is stopped.
+	//    When the gRPC connection is terminated on exiting Run, sendActionResults exits too.
+	// 3. processChannelRequests reads requests from the channel and processes them.
 	//    It exits when an unexpected message is received from the channel, or when can't be received at all.
-	//    When Run is left, caller stops supervisor, and that allows processSupervisorRequests to exit.
-	// Done() channel is closed when both goroutines exited.
-	oneDone := make(chan struct{}, 2)
+	//    When Run is left, caller stops supervisor FIXME BUT DOES NOT STOP ACTION RUNNER, and that allows processSupervisorRequests to exit.
```
AlekSi: There is a data race due to this.

client/client.go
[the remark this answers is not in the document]
AlekSi: >Done() channel is closed when *both* goroutines exited.

We start *three* goroutines.

client/client.go
DexterHD: > There is a data race due to this.

So the data race is because Run doesn't stop ActionRunner? Am I understand right?
And I can check this with the race detector?

I thought all work fine as `ActionRunner` stops all actions and closes the channel when `applicationContext` is done.

Here I'm close the channel.
https://github.com/percona/pmm-agent/blob/d02139b4dfcb1683ee8952018b1c9497faaebc8e/actions/concurrent_runner.go#L66-L69

And here I pass the application context to all actions.
https://github.com/percona/pmm-agent/blob/d02139b4dfcb1683ee8952018b1c9497faaebc8e/actions/concurrent_runner.go#L96-L104

What am I doing wrong here?

AlekSi: https://github.com/percona/pmm-agent/pull/52

actions/actions.go
```diff
@@ -0,0 +1,34 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+// Package actionsCancel provides PMM Actions implementations and related utils.
```
golangcibot: package comment should be of the form "Package actions ..." (from `golint`)

client/client.go
DexterHD: Fixed in https://github.com/percona/pmm-agent/pull/47/commits/00cf6f21df6e93b3342fb1f05ca1d77936e760ae

actions/process_action.go
```diff
@@ -0,0 +1,62 @@
...
+	id      string
+	command string
+	arg     []string
+}
+
+// NewProcessAction creates Shell action.
+//
+// Process action, it's an abstract action that can run an external commands.
+// This commands can be a shell script, script written on interpreted language, or binary file.
+func NewProcessAction(id string, cmd string, arg []string) action {
```
golangcibot: exported func NewProcessAction returns unexported type actions.action, which can be annoying to use (from `golint`)

DexterHD: @AlekSi Fix all, but now have this and also have the same issue in my IDE. :(

AlekSi: Ok, I will rename it back and merge this PR today.

actions/concurrent_runner.go
```diff
@@ -0,0 +1,122 @@
...
+		r.out <- actionResult{
+			ID:             a.ID(),
+			Error:          err,
+			CombinedOutput: cOut,
+		}
+	}()
+}
+
+// ActionReady returns channel that you can use to read Action results.
+func (r *ConcurrentRunner) ActionReady() <-chan actionResult {
```
golangcibot: exported method ActionReady returns unexported type <-chan actions.actionResult, which can be annoying to use (from `golint`)

DexterHD: 😭

PR #48, opened 2019-05-10
PMM-4007 Add --force to setup command.

AlekSi:


AlekSi:
Merged to unblock next work.

PR #53, opened 2019-05-16
PMM-4044: Proposal, leave interfaces in main file of the package.

DexterHD:
Is any reason we separate interfaces in `mocks.go` file? Maybe it will be better to leave interface declaration in the main file of the package as it uses in a lot of go projects? I think `make gen` will do the same, but the code will be a little cleaner, as these interfaces using not only for testing.

@AlekSi @BupycHuk What do you think? I'm not sure but it seems looks like more idiomatic.


BupycHuk:
I agree that it's better to keep interfaces in main file instead of mock file.

PR #54, opened 2019-05-16
PMM-4044 Proposal to shared type of client.

DexterHD:
@AlekSi @BupycHuk Maybe it would be better to move `NetworkInformation` inside `client` package, as this information generates by the client. To avoid dependency circle we can make subpackage. I saw this approach in a lot of projects. One problem is that our `main_test.go` prohibits not only package but all sub-packages. 😞


AlekSi:
Not sure if I like the idea of having the package `X/Y` with more visibility than package `X`. Maybe a better large example will be more convincing. But doing that for two `time.Duration` values do not make much sense to me.

DexterHD:
@AlekSi It's ok, I agree it doesn't make sense in a small example like this, but in general, it's ok.
In this case, I use `X` not as a package but as a `Domain` wich `NetworkInfo`  belongs to.

For example packages like `net/http` from the standard library. `net` is a domain, the same thing you can find in `html/template`. `html` is domain too. And there can be a lot of examples that it's fine.

PR #55, opened 2019-05-16
PMM-4044 Proposal for default dial timeout.

DexterHD:
@AlekSi @BupycHuk Not sure about idiomacity, but maybe this version of API could be cleaner? And I'm sorry, it seems this lines of code is not related to this ticket. So I think it can be closed.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -43,6 +43,7 @@ import (
 )
 
 const (
+	// DefaultDialTimeout is default timeout for client dial.
 	DefaultDialTimeout = 5 * time.Second
 	backoffMinDelay    = 1 * time.Second
 	backoffMaxDelay    = 15 * time.Second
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -42,21 +42,19 @@ import (
...
 
-var (
-	dialTimeout = 5 * time.Second // changed by unit tests
-)
-
 const (
-	backoffMinDelay   = 1 * time.Second
-	backoffMaxDelay   = 15 * time.Second
-	clockDriftWarning = 5 * time.Second
+	DefaultDialTimeout = 5 * time.Second
```
golangcibot: exported const DefaultDialTimeout should have comment (or a comment on this block) or be unexported (from `golint`)

AlekSi:
The upside for this PR is that it replaces variable with a constant.
The downside is that it exports more configuration to the caller. I don't think this is a good thing in many cases, including this one.
See 56d26f664912f26dee17e52f9d447b501c3ae228.

PR #77, opened 2019-06-17
PMM-4201 Fix some fluky tests.

AlekSi:


actions/mysql_show_index_action_test.go
```diff
@@ -68,6 +67,7 @@ func TestShowIndex(t *testing.T) {
 			assert.Equal(t, []interface{}{"city", "1", "CountryCode", "1", "CountryCode", "A", "465", nil, nil, "", "BTREE", "", ""}, actual[2])
 
 		case mySQLVersion == "5.7":
+			actual[2][6] = "232"
```
BupycHuk: is that value random one?

actions/concurrent_runner.go
```diff
@@ -31,7 +31,7 @@ const defaultTimeout = time.Second * 10
 type ActionResult struct {
 	ID     string
 	Output []byte
-	Error  error
+	Error  string
```
BupycHuk: Why did you decide to use string instead of error?

AlekSi: Mainly to simplify test – `signal: killed` error is an unexported type.
We converted it to `string` in the caller, so nothing changed in the end.

client/client.go
```diff
@@ -199,16 +199,11 @@ func (c *Client) Done() <-chan struct{} {
 
 func (c *Client) processActionResults() {
 	for result := range c.runner.Results() {
-		var errMessage string
-		if result.Error != nil {
-			errMessage = result.Error.Error()
```
AlekSi: @BupycHuk there

BupycHuk: oh, ok

actions/mysql_show_index_action_test.go
AlekSi: `ANALYZE TABLE` (that we call in `OpenTestMySQL`) is supposed to update table statistics, so tests should be stable. But they are not in fact 100% stable. That value is one that is produced by about 95% of the runs.

> // cardinality sometimes changes between runs; fix it to the most common value for that version

AlekSi:
Tests pass. 🎉

PR #81, opened 2019-06-21
PMM-4214 Add flag to skip registration on PMM Server.

AlekSi:


AlekSi:
Another fluky test added to the black notebook. 😆

PR #82, opened 2019-06-24
PMM-4235 Temporary remove github.com/lfittl/pg_query_go.

AlekSi:


AlekSi:
That was a trigger, but not the cause.

PR #86, opened 2019-07-01
PMM-4278 QAN doesn't display data for MongoDB.

DexterHD:


Gopkg.toml
```diff
@@ -49,7 +49,7 @@ required = [
 # We need https://github.com/percona/percona-toolkit/pull/396.
 [[constraint]]
   name = "github.com/percona/percona-toolkit"
-  branch = "3.0"
+  branch = "PMM-4278-official-mongo-driver"
```
AlekSi: Why there is no PR for that branch?

agents/mongodb/internal/profiler/collector/collector.go
```diff
@@ -180,11 +180,13 @@ func start(wg *sync.WaitGroup, client *mongo.Client, dbName string, docsChan cha
 	// signal WaitGroup when goroutine finished
 	defer wg.Done()
 
+	timeoutCtx, cancel := context.WithTimeout(context.TODO(), time.Second*5)
```
AlekSi: `5*time.Second`

Gopkg.toml
DexterHD: Because we can't merge it. We are depending on https://github.com/percona/percona-toolkit/tree/PMM-4278-official-mongo-driver/src/go/mongolib, and it uses `mgo.v2/bson`. 

As we merged to `mongo-driver` we can't use `mgo.v2/bson` and should use an official one. 

So we talk with @rnovikovP and decide to make the different branch and change `mgo.v2/bson` to `mongo-driver/bson` to make this works.

rnovikovP: we also should  change  https://github.com/Percona-Lab/pmm-submodules/blob/PMM-2.0/.gitmodules#L42 for  beta3

PT migration will be tracked in a separate issue

AlekSi: >> Why there is no PR for that branch?

> Because we can't merge it.

But it doesn't mean we can't create it. Tracking PRs is easier.

AlekSi:
https://github.com/percona/percona-toolkit/pull/404 Test failures look scary.

DexterHD:
> [percona/percona-toolkit#404](https://github.com/percona/percona-toolkit/pull/404) Test failures look scary.

Yep. 😕  But I have changed only `mgo.v2/bson` to `mongo-driver/bson`. All test depends on `mgo.v2`. To make it work, it should be changed to `mongo-driver`. There was no such task like "Migrate percona-toolkit to mongo-driver". I've solved the concrete bug. All next decisions are up to you and @rnovikovP . No offense.

PR #90, opened 2019-07-11
 PMM-4308 Fix slowlog test on CI.

BupycHuk:


.travis.yml
```diff
@@ -47,7 +47,12 @@ env:
     - MYSQL_IMAGE=mariadb:10.4
 
 before_install:
+  - touch testdata/slowlogs/slow.log
+  - chmod 766 testdata/slowlogs/slow.log
```
AlekSi: Why 766 and not 666? I'm not sure why we want it to br executable.

.travis.yml
```diff
@@ -47,7 +47,12 @@ env:
...
   - docker-compose up -d
+  - sleep 20
+  - docker-compose logs
+  - ls -la testdata/slowlogs/
```
AlekSi: Let's remove those debug commands.

PR #126, opened 2019-09-27
PMM-4757 Use unsigned ints.

AlekSi:


agents/mysql/perfschema/perfschema.go
```diff
@@ -237,7 +245,10 @@ func makeBuckets(current, prev map[string]*eventsStatementsSummaryByDigest, l *l
 		if prevESS == nil {
 			prevESS = new(eventsStatementsSummaryByDigest)
 		}
-		count := float32(currentESS.CountStar - prevESS.CountStar)
+
+		// convert to (signed) float64 to avoid uint64 wrap-around
+		count := float32(float64(currentESS.CountStar) - float64(prevESS.CountStar))
```
BupycHuk: why don't to compare these two values in `switch-case` below? and set count value right after `switch-case` statement using method `inc`.

BupycHuk: since currentESS.CountStar is uint64 then https://github.com/percona/pmm-agent/pull/126/files#diff-f7f27c973c5ca602e5dad3b3a8760cbcR264 might work incorrectly.

AlekSi: Not sure I got your second comment

PR #136, opened 2019-11-21
PMM-4966 Filter sensitive information

BupycHuk:


agents/process/process_logger.go
```diff
@@ -88,6 +92,13 @@ func (pl *processLogger) Latest() []string {
 	return result
 }
 
+func (pl *processLogger) clearKeywords(s string) string {
+	for _, k := range pl.hideKeywords {
+		s = strings.Replace(s, k, "***", -1)
+	}
+	return s
```
golangcibot: return statements should not be cuddled if block has more than two lines (from `wsl`)

agents/process/process_logger.go
```diff
@@ -64,6 +67,7 @@ func (pl *processLogger) Write(p []byte) (n int, err error) {
 			return
 		}
 		line = strings.TrimSuffix(line, "\n")
+		line = pl.clearKeywords(line)
```
askomorokhov: AFAIK, in repeated usage, a Replacer is faster than many calls to Replace.
So maybe create Replacer outside of `for {}` will be better here.

Makefile
```diff
@@ -56,7 +56,7 @@ install-race:                   ## Install pmm-agent binary with race detector.
 TEST_PACKAGES ?= ./...
 TEST_FLAGS ?= -timeout=20s
 
-test:                           ## Run tests.
+test: install                   ## Run tests.
```
AlekSi: I specifically removed it – tests are cached, but the final binary linking is not, and is is super-slow on macOS due to cgo.

Makefile
```diff
@@ -118,5 +118,8 @@ env-up:                         ## Start development environment.
 env-down:                       ## Stop development environment.
 	docker-compose down --volumes --remove-orphans
 
+setup-dev:
```
AlekSi: Here I think we should depend on `install` and use `pmm-agent`. Please add a doc comment – it is visible on `make` without arguments.

agents/process/process_logger.go
```diff
@@ -88,6 +92,13 @@ func (pl *processLogger) Latest() []string {
 	return result
 }
 
+func (pl *processLogger) clearKeywords(s string) string {
+	for _, k := range pl.hideKeywords {
+		s = strings.Replace(s, k, "***", -1)
```
AlekSi: > Replace returns a copy of the string s with the first n non-overlapping instances of old replaced by new. **If old is empty, it matches at the beginning of the string and after each UTF-8 sequence, yielding up to k+1 replacements for a k-rune string.**

That's very counter-intuitive when you encounter it. Let's skip, or log and skip, or even panic if `k == ""`.

BupycHuk: ok

Makefile
BupycHuk: ok

Makefile
BupycHuk: ok

Makefile
BupycHuk: why do we should depend on install?

AlekSi: > and use pmm-agent

To run the latest installed pmm-agent – to avoid ldflags copy&pasting.

BupycHuk: but I run it by `go run main.go` here.

AlekSi: > to avoid ldflags copy&pasting.

AlekSi: Look, I don't mind much. I just don't see a competing reason to make it different from `run` / `run-race`.

BupycHuk: ah, ok, got you

agents/process/process_logger.go
```diff
@@ -88,6 +94,17 @@ func (pl *processLogger) Latest() []string {
 	return result
 }
 
+func (pl *processLogger) replacer() *strings.Replacer {
+	var r []string
```
golangcibot: Consider preallocating `r` (from `prealloc`)

agents/process/process_logger.go
```diff
@@ -88,6 +94,17 @@ func (pl *processLogger) Latest() []string {
...
+	for _, k := range pl.redactWords {
+		if k == "" {
+			panic("redact word can't be empty")
+		}
+		r = append(r, k, "***")
```
golangcibot: append only allowed to cuddle with appended value (from `wsl`)

agents/process/process_logger.go
```diff
@@ -88,6 +94,17 @@ func (pl *processLogger) Latest() []string {
 	return result
 }
 
+func (pl *processLogger) replacer() *strings.Replacer {
+	var r []string
+	for _, k := range pl.redactWords {
```
golangcibot: ranges should only be cuddled with assignments used in the iteration (from `wsl`)

agents/process/process_logger.go
```diff
@@ -88,6 +94,17 @@ func (pl *processLogger) Latest() []string {
...
+		if k == "" {
+			panic("redact word can't be empty")
+		}
+		r = append(r, k, "***")
+	}
+	return strings.NewReplacer(r...)
```
golangcibot: return statements should not be cuddled if block has more than two lines (from `wsl`)

agents/process/process_logger.go
AlekSi: TIL, thanks for sharing

agents/process/process_logger.go
```diff
@@ -64,6 +74,9 @@ func (pl *processLogger) Write(p []byte) (n int, err error) {
 			return
 		}
 		line = strings.TrimSuffix(line, "\n")
+		if pl.replacer != nil {
```
golangcibot: only one cuddle assignment allowed before if statement (from `wsl`)

agents/process/process_logger.go
```diff
@@ -88,6 +101,17 @@ func (pl *processLogger) Latest() []string {
 	return result
 }
 
+func replacer(redactWords []string) *strings.Replacer {
+	r := make([]string, 0, len(redactWords)*2)
+	for _, w := range redactWords {
+		if w == "" {
+			panic("redact word can't be empty")
+		}
+		r = append(r, w, "***")
```
golangcibot: append only allowed to cuddle with appended value (from `wsl`)

agents/process/process_logger.go
```diff
@@ -88,6 +101,17 @@ func (pl *processLogger) Latest() []string {
 	return result
 }
 
+func replacer(redactWords []string) *strings.Replacer {
+	r := make([]string, 0, len(redactWords)*2)
+	for _, w := range redactWords {
```
golangcibot: ranges should only be cuddled with assignments used in the iteration (from `wsl`)

PR #140, opened 2019-12-10
PMM-4927 Fix test failures.

BupycHuk:


agents/mysql/perfschema/perfschema_test.go
```diff
@@ -234,6 +234,7 @@ func TestPerfSchema(t *testing.T) {
 	require.NoError(t, err)
 	tests.LogTable(t, structs)
 
+	rowsExamined := float32(0)
```
AlekSi:
```go
var rowsExamined float32
```
Use that syntax for zero values.

BupycHuk:
@AlekSi I tried to remove `filter` function, but it makes tests unstable. We run setup each time and truncate perfschema table, but it doesn't help. So I would prefer to keep `filter` function. Btw I have no idea how filter could help with parallel testing issue.

PR #154, opened 2020-04-17
SAAS-20 Query API.

AlekSi:


actions/postgresql_query_show_action_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"time"
+
+	"github.com/percona/pmm/api/agentpb"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/percona/pmm-agent/utils/tests"
+)
+
+func TestPostgreSQLQueryShow(t *testing.T) {
```
golangcibot, line 30: func `TestPostgreSQLQueryShow` is unused (from `unused`)

actions/mysql_query_select_action.go
```diff
@@ -0,0 +1,75 @@
...
+// Run runs an Action and returns output and error.
+func (a *mysqlQuerySelectAction) Run(ctx context.Context) ([]byte, error) {
+	db, err := mysqlOpen(a.params.Dsn)
+	if err != nil {
+		return nil, err
+	}
+	defer db.Close() //nolint:errcheck
+
+	// use prepared statement to force binary protocol usage that returns correct types
+	stmt, err := db.PrepareContext(ctx, "SELECT /* pmm-agent */ "+a.params.Query) //nolint:gosec
```
artemgavrilov, line 57: We can also add validation that the query doesn't contain a semicolon, to prevent injections. Also probably we can forbid UNION, DROP, TRUNCATE, and other keywords that obviously out of the scope of STT. Maybe it should be done in SAAS repo.

actions/postgresql_query_select_action.go
```diff
@@ -0,0 +1,82 @@
...
+	db := sql.OpenDB(connector)
+	defer db.Close() //nolint:errcheck
+
+	// A very basic check that there is a single SELECT query. It has oblivious false positives (`SELECT ';'`),
+	// but PostgreSQL query lexical structure (https://www.postgresql.org/docs/current/sql-syntax-lexical.html)
+	// does not allow false negatives.
+	// If we decide to improve it, we could use our existing query parser from pg_stat_statement agent,
+	// or use a simple hand-made parser similar to
+	// https://github.com/mc2soft/pq-types/blob/ada769d4011a027a5385b9c4e47976fe327350a6/string_array.go#L82-L116
+	if strings.Contains(a.params.Query, ";") {
```
artemgavrilov, line 66: Oh, I found such check)

AlekSi: In MySQL it is handled by the driver setting without false positives: https://github.com/percona/pmm-managed/blob/4409b0a0a8c69420f82775893e9aa453da911a88/models/agent_model.go#L192-L193 And there are lot of tests will be broken if we accidentally change it.

actions/mysql_query_select_action.go
AlekSi: You should not be able to write a query like `SELECT … UNION DROP/TRUNCATE …`. But please try it. If one can do that, then we should definitely do something about it.

PR #155, opened 2020-04-22
Enable debug/trace logging with PMM_DEBUG/TRACE.

AlekSi:
Those environment variables enable debug/trace logging in all PMM
components. That's mostly useful for PMM Server in Docker:
$ docker run -e PMM_DEBUG=1 ...

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/percona/pmm/utils/nodeinfo"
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -213,6 +214,14 @@ func get(args []string, l *logrus.Entry) (cfg *Config, configFileF string, err e
...
 			}
 		}
+
+		// enabled cross-component PMM_DEBUG and PMM_TRACE take priority
+		if b, _ := strconv.ParseBool(os.Getenv("PMM_DEBUG")); b {
+			cfg.Debug = true
+		}
+		if b, _ := strconv.ParseBool(os.Getenv("PMM_TRACE")); b {
+			cfg.Trace = true
+		}
```
artemgavrilov, line 224: Can be easier:
```suggestion
		cfg.Debug, _ = strconv.ParseBool(os.Getenv("PMM_DEBUG"))
		cfg.Trace, _ = strconv.ParseBool(os.Getenv("PMM_TRACE"))
```

artemgavrilov: Hm, or can't

AlekSi: Yeah, it will disable debug logging if PMM_DEBUG is not set, but PMM_AGENT_DEBUG is set to 1.

PR #157, opened 2020-05-07
Use standard port for PostgreSQL.

AlekSi:
There is no port clash with pmm-managed anymore because it uses devcontainer.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,7 +57,7 @@ services:
       -c pg_stat_statements.track=all
       -c pg_stat_statements.save=off
     ports:
-      - 127.0.0.1:15432:5432
+      - 127.0.0.1:5432:5432
     environment:
       - POSTGRES_USER=pmm-agent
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -57,7 +57,7 @@ services:
       -c pg_stat_statements.track=all
       -c pg_stat_statements.save=off
     ports:
-      - 127.0.0.1:15432:5432
+      - 127.0.0.1:5432:5432
```
BupycHuk, line 60: it was done because it was conflicting on my local system

BupycHuk: with local installed postgresql

AlekSi: But do you use it?

My motivation was to make it behave the same as MySQL and MongoDB for local checks development:
```
pmm-admin add mysql
pmm-admin add mongodb
pmm-admin add postgresql postgresql 127.0.0.1:15432
```

AlekSi: @BupycHuk ^

PR #162, opened 2020-05-19
PMM-5723 Reviewdog checks.

BupycHuk:


.golangci.yml
```diff
@@ -32,10 +32,6 @@ linters:
     - wsl               # too annoying
     - gomodguard        # we are not using modules
 
-run:
-  skip-dirs:
-    - agents/mongodb/internal/
```
AlekSi, line 37: We still have that directory, and the code there still will not pass linters.

Makefile
```diff
@@ -155,3 +155,10 @@ env-sysbench-run:
 		--db-driver=pgsql --pgsql-host=postgres --pgsql-user=pmm-agent --pgsql-password=pmm-agent-password --pgsql-db=pmm-agent \
 		--threads=4 --time=0 --rate=10 --report-interval=10 --percentile=99 \
 		--tables=1 --scale=10  --use_fk=0 --enable_purge=yes run
+
+check-all: check        ## Run golang ci linter to check new changes from master.
```
AlekSi, line 159: That should follow `check` target, and doc comments should be indented with the same number of spaces.
Basically, port https://github.com/percona/pmm-managed/commit/cf59a2d57f6c2b966160c9a959a4a3fdd780cafb. Please do the same for other PRs.

.golangci.yml
BupycHuk: Yeah, but it's not synced with upstream code, so we should check it each time we update code.

AlekSi: I think you confuse it with Prometheus code what we had in pmm-managed.
This is our own code.

BupycHuk: So, what's the reason to ignore it?

AlekSi: It was copied from qan-agent as is. It can't be improved, it should be rewritten. And running golangci-lint on that directory produces so many warnings that it hides all other problems.

BupycHuk: We run golangci-lint only for diff between current branch and master, so mostly it won't be visible.

AlekSi: As I said before – please run linter locally. This file contains configuration for local linter too.

BupycHuk: https://github.com/percona/pmm-agent/pull/162/files#diff-b67911656ef5d18c4ae36cb6741b7965R160

PR #163, opened 2020-05-21
PMM-5839 Fixed incorrect metrics collection.

BupycHuk:


BupycHuk:
@AlekSi It's not possible to write separate test with a creating mock table with data, selecting data and checking m_blk_read_time. Postgres keeps data in memory and doesn't read from the disk, that's why block read time is equal to zero.

AlekSi:
@BupycHuk CI fails

agents/postgres/pgstatstatements/pgstatstatements.go
```diff
@@ -242,8 +242,8 @@ func makeBuckets(current, prev map[int64]*pgStatStatementsExtended, l *logrus.En
 			{float32(currentPSS.TempBlksRead - prevPSS.TempBlksRead), &mb.Postgresql.MTempBlksReadSum, &mb.Postgresql.MTempBlksReadCnt},
 			{float32(currentPSS.TempBlksWritten - prevPSS.TempBlksWritten), &mb.Postgresql.MTempBlksWrittenSum, &mb.Postgresql.MTempBlksWrittenCnt},
 
-			{float32(currentPSS.BlkReadTime - prevPSS.BlkReadTime), &mb.Postgresql.MBlkReadTimeSum, &mb.Postgresql.MBlkReadTimeCnt},
-			{float32(currentPSS.BlkWriteTime - prevPSS.BlkWriteTime), &mb.Postgresql.MBlkWriteTimeSum, &mb.Postgresql.MBlkWriteTimeCnt},
+			{float32(currentPSS.BlkReadTime-prevPSS.BlkReadTime) / 1000, &mb.Postgresql.MBlkReadTimeSum, &mb.Postgresql.MBlkReadTimeCnt},
```
AlekSi, line 245: Let's add a comment as above:
```
// convert milliseconds to seconds
```

AlekSi:
@BupycHuk I think we should tweak tests, not just restart them

BupycHuk:
I'm trying to figure out is this because of changes or it's just unstable. Locally it works fine.

AlekSi:
This is similar to how our QAN data collection tests failed initially – we want them to be precise as much as possible, but they can't be fully precise by nature. We don't want to completely skip size checks and don't want to use a too wide range, but we have to tweak every now and then for some time.

actions/postgresql_query_show_action_test.go
```diff
@@ -45,13 +45,13 @@ func TestPostgreSQLQueryShow(t *testing.T) {
 
 		b, err := a.Run(ctx)
 		require.NoError(t, err)
-		assert.LessOrEqual(t, 22151, len(b))
+		assert.GreaterOrEqual(t, len(b), 22150)
 		assert.LessOrEqual(t, len(b), 29156)
```
AlekSi, line 49: It was done intentionally: `22151 < len(b) < 29156`

AlekSi:
Does discussion in Slack change anything?

actions/postgresql_query_show_action_test.go
BupycHuk: And it makes it more clear to understand that len(b) should be greater then 22150 and less then 29156

BupycHuk:
@AlekSi yes, I figured out how to make a test

actions/postgresql_query_show_action_test.go
AlekSi: Well, I don't agree that it makes it more clear: I read origin code as `22151 <= len(b) <= 29156` (as one would write in Python or just in math) and new code as `len(b) >= 22150 && len(b) <= 29156`.

But if you insist, please change that in other files and tests.

agents/postgres/pgstatstatements/pgstatstatements_test.go
```diff
@@ -295,5 +309,86 @@ func TestPGStatStatementsQAN(t *testing.T) {
...
+	t.Run("CheckMBlkReadTime", func(t *testing.T) {
+		r := rand.New(rand.NewSource(time.Now().Unix()))
+		tableName := fmt.Sprintf("customer%d", r.Int())
+		_, err := db.Exec(fmt.Sprintf(`
+		CREATE TABLE %s (
+			customer_id integer NOT NULL,
+			first_name character varying(45) NOT NULL,
+			last_name character varying(45) NOT NULL,
+			active boolean
+		);`, tableName))
```
AlekSi, line 324: There is no real need to add `;` to queries

agents/postgres/pgstatstatements/pgstatstatements_test.go
```diff
@@ -295,5 +309,86 @@ func TestPGStatStatementsQAN(t *testing.T) {
...
+		require.NoError(t, err)
+		defer func() {
+			_, err := db.Exec(fmt.Sprintf(`DROP TABLE %s;`, tableName))
+			require.NoError(t, err)
+		}()
+		m := setup(t, db)
+
+		waitGroup := sync.WaitGroup{}
```
AlekSi, line 332: That's something I say over and over again: there is no field you can possible initialise in WaitGroup, so please use `var wg sync.WaitGroup`.

And I think we should add go-consistent to this repo.

actions/postgresql_query_show_action_test.go
BupycHuk: Ok, reverted it.

agents/postgres/pgstatstatements/pgstatstatements_test.go
BupycHuk: I'll create task for go-consistent

agents/postgres/pgstatstatements/pgstatstatements_test.go
```diff
@@ -295,5 +309,86 @@ func TestPGStatStatementsQAN(t *testing.T) {
...
+		_, err := db.Exec(fmt.Sprintf(`
+		CREATE TABLE %s (
+			customer_id integer NOT NULL,
+			first_name character varying(45) NOT NULL,
+			last_name character varying(45) NOT NULL,
+			active boolean
+		)`, tableName))
+		require.NoError(t, err)
+		defer func() {
+			_, err := db.Exec(fmt.Sprintf(`DROP TABLE %s;`, tableName))
```
AlekSi, line 327: `;`

PR #165, opened 2020-05-28
PMM-5365 Fix UTF-8 problem.

JiriCtvrtka:
- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/892


BupycHuk:
looks like wrong issue number in PR header.

JiriCtvrtka:
@BupycHuk Yes, changed, thank you.

agents/mysql/slowlog/slowlog.go
```diff
@@ -383,7 +384,7 @@ func makeBuckets(agentID string, res event.Result, periodStart time.Time, period
 		}
 
 		if v.Example != nil && !disableQueryExamples {
-			mb.Common.Example = v.Example.Query
+			mb.Common.Example = strings.ToValidUTF8(v.Example.Query, "")
```
BupycHuk, line 387: Can we add test for it?

agents/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -251,13 +252,15 @@ func (a *Aggregator) createResult(ctx context.Context) *report.Result {
 		s := strings.SplitN(v.Namespace, ".", 2)
 		if len(s) == 2 {
 			db = s[0]
-			collection = s[1]
+			collection, _ = truncate.Query(s[1])
 		}
 
+		example, _ := truncate.Query(v.Query)
```
BupycHuk, line 258: the same here

agents/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -251,13 +252,15 @@ func (a *Aggregator) createResult(ctx context.Context) *report.Result {
...
+		fingerprint, _ := truncate.Query(v.Fingerprint)
 		bucket := &agentpb.MetricsBucket{
 			Common: &agentpb.MetricsBucket_Common{
```
BupycHuk, line 261: we have isTruncated here and it should be used here.

BupycHuk: if fingerprint or example is truncated `isTruncated` should be true.

BupycHuk: Here and in other agents

agents/postgres/pgstatstatements/pgstatstatements.go
```diff
@@ -206,13 +207,14 @@ func makeBuckets(current, prev map[int64]*pgStatStatementsExtended, l *logrus.En
 			l.Debugf("Normal query: %s.", currentPSS)
 		}
 
+		fingerprint, _ := truncate.Query(currentPSS.Query)
```
BupycHuk, line 210: it's already truncated here https://github.com/percona/pmm-agent/blob/2c5ff8cc17b0416f770daa9cb37d99bc0e55f04a/agents/postgres/pgstatstatements/stat_statements_cache.go#L140

agents/mongodb/internal/profiler/aggregator/aggregator.go
BupycHuk: Why did you resolve this? I see that we still don't use isTruncated

agents/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -99,6 +100,9 @@ func (a *Aggregator) Add(ctx context.Context, doc proto.SystemProfile) error {
 	// we had some activity so reset timer
 	a.t.Reset(a.d)
 
+	doc.Ns, _ = truncate.Query(doc.Ns)
```
BupycHuk, line 103: Any reason to call it here?

agents/mongodb/internal/profiler/aggregator/aggregator.go
JiriCtvrtka: I resolve it yesterday after I pushed commit "PMM-5365 Patch." because truncating was not in this commit.

agents/mongodb/internal/profiler/aggregator/aggregator.go
JiriCtvrtka: From tests I found out that is needed here too. Because from Ns and Op it assembly Fingerprint and Example.

agents/mongodb/internal/profiler/aggregator/aggregator.go
JiriCtvrtka: @BupycHuk Just some clarification. If example return truncated true and fingerprint false, what is right behavior? If any will be true then I always return true?

BupycHuk: yes, if any of them is true then it should be true

agents/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -99,6 +99,9 @@ func (a *Aggregator) Add(ctx context.Context, doc proto.SystemProfile) error {
 	// we had some activity so reset timer
 	a.t.Reset(a.d)
 
+	doc.Ns = strings.ToValidUTF8(doc.Ns, "")
```
BupycHuk, line 102: please add a comment why we are doing it.

agents/mysql/perfschema/perfschema.go
```diff
@@ -266,16 +267,22 @@ func makeBuckets(current, prev map[string]*eventsStatementsSummaryByDigest, l *l
 			l.Debugf("Normal query: %s.", currentESS)
 		}
 
+		var isTruncated bool
+		fingerprint, truncated := truncate.Query(*currentESS.DigestText)
+		if truncated {
```
BupycHuk, line 272: i think we don't need this if here

agents/mysql/slowlog/slowlog.go
```diff
@@ -363,10 +364,15 @@ func makeBuckets(agentID string, res event.Result, periodStart time.Time, period
 			continue
 		}
 
+		var isTruncated bool
```
BupycHuk, line 367: looks like this extra variable is not required, we can use truncated itself

utils/truncate/query_test.go
```diff
@@ -38,8 +38,8 @@ func TestQuery(t *testing.T) {
 		"абвгдеё": {"а ...", true},
 
 		// Unicode replacement characters
```
BupycHuk, line 40: please add comment here, why \uFFFD doesn't work for us.

agents/mysql/perfschema/perfschema.go
BupycHuk: we can use `truncated` instead of `isTruncated`

agents/mysql/slowlog/slowlog_test.go
```diff
@@ -41,6 +42,46 @@ func getDataFromFile(t *testing.T, filePath string, data interface{}) {
...
+	const agentID = "/agent_id/73ee2f92-d5aa-45f0-8b09-6d3df605fd44"
+	periodStart := time.Unix(1557137220, 0)
+
+	parsingResult := event.Result{
+		Class: map[string]*event.Class{
+			"example": {
+				Metrics:     &event.Metrics{},
+				Fingerprint: "set lock_wait_timeout=?\xff",
+				Example: &event.Example{
+					Query: "ping \xff",
```
askomorokhov, line 55: Maybe fingerprint and query should be the same. (query - data = fingerprint)

agents/mongodb/internal/profiler/aggregator/aggregator.go
BupycHuk: I mean that we have issue with converting invalid symbols to valid utf-8 symbols like `\uFFFD` and that we have issue with sending this data.

utils/truncate/query_test.go
AlekSi: Yeah, I wonder why it doesn't work for you.

`[]rune(string)` converts a binary string to Unicode runes, replacing each sequence of bytes that can't be a valid Unicode rune with the [Unicode replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character) (which itself is a valid Unicode rune). That what this function did, that's what the comment below means, and that's why this behavior was tested. Those are not "invalid chars".

So I wonder why you decided to strip them completely instead of replacing them by a rune specifically designed for that case.

agents/mysql/slowlog/slowlog_test.go
```diff
@@ -41,6 +42,46 @@ func getDataFromFile(t *testing.T, filePath string, data interface{}) {
...
 
+func TestSlowLogMakeBucketsInvalidUTF8(t *testing.T) {
+	const agentID = "/agent_id/73ee2f92-d5aa-45f0-8b09-6d3df605fd44"
+	periodStart := time.Unix(1557137220, 0)
+
+	parsingResult := event.Result{
+		Class: map[string]*event.Class{
+			"example": {
+				Metrics:     &event.Metrics{},
+				Fingerprint: "set lock_wait_timeout=?\xff",
```
AlekSi, line 53: Can we make a somewhat more real test? That show what data we lose from the real requests.

agents/mongodb/internal/profiler/aggregator/aggregator_test.go
```diff
@@ -94,4 +96,56 @@ func TestAggregator(t *testing.T) {
...
+		agentID := "test-agent"
+		startPeriod := time.Now()
+		aggregator := New(startPeriod, agentID, logrus.WithField("component", "test"))
+		aggregator.Start()
+		defer aggregator.Stop()
+		ctx := context.TODO()
+		err := aggregator.Add(ctx, proto.SystemProfile{
+			NscannedObjects: 2,
+			Nreturned:       3,
+			Ns:              "collection.people\xff",
```
AlekSi, line 110: Can we make a somewhat more real test?

utils/truncate/query_test.go
JiriCtvrtka: When it was replaced by \uFFFD it still caused issue with invalid UTF-8 char, even it is valid UTF-8 char. Maybe it is problem of conversion in diffrent location.

AlekSi: Please make a test that shows that encoding U+FFFD in protobuf string produces an error

BupycHuk: AFAIK, even with these characters we are still having the same error.

JiriCtvrtka: I am not sure if I can do it in tests. To test it I always release binary and then test it this way. Here are steps which are needed to reproduce it. 

Maybe, this is similar problem even with valid UTF-8 chars:
https://bugs.mysql.com/bug.php?id=78758

AlekSi: > Here are steps which are needed to reproduce it.

Where?

MySQL bug doesn't involve protobuf, so I'm not sure how that can be related.

I'm sure you can write a test that shows that protobuf can't encode U+FFFD – just encode that string with protobuf and check for error. That's less than 10 lines.

JiriCtvrtka: @AlekSi Test passed, also I tried again binary without ToValidUTF8 (only with truncate) method and it is fine. For first time I used probably wrong binary. So you was right and it is not necessary. I will remove it and modify tests for it.

AlekSi: > Test passed

Please add it there

agents/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -254,10 +255,23 @@ func (a *Aggregator) createResult(ctx context.Context) *report.Result {
...
+		fingerprint, truncated := truncate.Query(v.Fingerprint)
+		if truncated {
+			isTruncated = truncated
+		}
+		query, truncated := truncate.Query(v.Query)
+		if truncated {
+			isTruncated = truncated
+		}
+		collection, truncated = truncate.Query(collection)
+		if truncated {
```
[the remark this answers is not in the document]
JiriCtvrtka: Yes, you are right.

agents/mongodb/internal/profiler/aggregator/aggregator_test.go
```diff
@@ -94,4 +97,71 @@ func TestAggregator(t *testing.T) {
...
+			Nreturned:       3,
+			Ns:              "collection." + table,
+			Op:              operation,
+		})
+		require.NoError(t, err)
+
+		result := aggregator.createResult(ctx)
+
+		var isTruncated bool
+		expectedTable, truncated := truncate.Query(table)
```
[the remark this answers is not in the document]
JiriCtvrtka: Changed back.

agents/mongodb/internal/profiler/aggregator/aggregator_test.go
```diff
@@ -94,4 +97,71 @@ func TestAggregator(t *testing.T) {
...
+		if truncated {
+			isTruncated = truncated
+		}
+		expectedOperation, truncated := truncate.Query(operation)
+		if truncated {
+			isTruncated = truncated
+		}
+
+		require.Equal(t, 1, len(result.Buckets))
+		assert.Equal(t, true, utf8.ValidString(result.Buckets[0].Common.Fingerprint))
```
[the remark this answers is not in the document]
JiriCtvrtka: Changed.

agents/mongodb/internal/profiler/aggregator/aggregator_test.go
```diff
@@ -94,4 +97,71 @@ func TestAggregator(t *testing.T) {
...
+		assert.Equal(t, true, utf8.ValidString(result.Buckets[0].Common.Example))
+		tests.AssertBucketsEqual(t, &agentpb.MetricsBucket{
+			Common: &agentpb.MetricsBucket_Common{
+				Queryid:             result.Buckets[0].Common.Queryid,
+				Fingerprint:         expectedOperation + " " + expectedTable,
```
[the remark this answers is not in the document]
JiriCtvrtka: Changed back.

agents/mysql/perfschema/perfschema_test.go
```diff
@@ -163,6 +164,40 @@ func TestPerfSchemaMakeBuckets(t *testing.T) {
...
+					NumQueries:  5,
+				},
+				Mysql: &agentpb.MetricsBucket_MySQL{
+					MRowsAffectedCnt: 5,
+					MRowsAffectedSum: 10,
+				},
+			},
+		}
+		require.Equal(t, 1, len(actual))
+		for i := range actual {
```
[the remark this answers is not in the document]
JiriCtvrtka: True, it is unnecessary.

utils/truncate/query_test.go
```diff
@@ -38,11 +40,23 @@ func TestQuery(t *testing.T) {
...
-		"\xff\xff\xff\xff\xff":     {"\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD", false},
-		"\xff\xff\xff\xff\xff\xff": {"\uFFFD ...", true},
+		"\xff\xff\xff\xff\xff":     {"\ufffd\ufffd\ufffd\ufffd\ufffd", false},
```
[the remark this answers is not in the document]
JiriCtvrtka: It was mistake, because I dont revert it, but write it back.

utils/truncate/query_test.go
```diff
@@ -38,11 +40,23 @@ func TestQuery(t *testing.T) {
...
+		"\xff\xff\xff\xff\xff\xff": {"\ufffd ...", true},
 	} {
 		query, truncated := Query(q)
 		assert.Equal(t, expected.query, query)
 		assert.Equal(t, expected.truncated, truncated)
 	}
 }
+
+func TestProtobuf(t *testing.T) {
+	query, _ := Query("SELECT * FROM contacts t0 WHERE t0.person_id = '߿�\xff\\uD83D\xdd'")
```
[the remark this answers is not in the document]
JiriCtvrtka: Ok.

client/channel/channel_test.go
```diff
@@ -91,6 +93,52 @@ func setup(t *testing.T, connect func(agentpb.Agent_ConnectServer) error, expect
 	return
 }
 
+func TestAgentRequestWithTruncatedInvalidUTF8(t *testing.T) {
+	const count = 50
```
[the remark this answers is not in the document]
JiriCtvrtka: Actually one is enough. Removed.

agents/mysql/perfschema/perfschema.go
```diff
@@ -266,16 +267,18 @@ func makeBuckets(current, prev map[string]*eventsStatementsSummaryByDigest, l *l
...
 			Common: &agentpb.MetricsBucket_Common{
 				Schema:                 pointer.GetString(currentESS.SchemaName), // TODO can it be NULL?
 				Queryid:                *currentESS.Digest,
-				Fingerprint:            *currentESS.DigestText,
+				Fingerprint:            fingerprint,
 				NumQueries:             count,
 				NumQueriesWithErrors:   inc(currentESS.SumErrors, prevESS.SumErrors),
 				NumQueriesWithWarnings: inc(currentESS.SumWarnings, prevESS.SumWarnings),
 				AgentType:              inventorypb.AgentType_QAN_MYSQL_PERFSCHEMA_AGENT,
+				IsTruncated:            truncated,
```
AlekSi, line 281: IsTruncated is documented as "Indicates if query examples is too long and was truncated" (with a grammar error).
The fingerprint is not a query example. Fingerprint also can't contain invalid UTF-8 bytes sequences – there are not allowed in table or columns names, and all values are absent in the fingerprint.

I think we should stay on the topic of this ticket and fix UTF-8 problems where they exist – in query examples.

(that's why I mentioned several times that we need realistic examples – it is not about replacing invalid UTF-8 bytes sequences in opaque strings, but where they are actually can be present)

agents/mongodb/internal/profiler/aggregator/aggregator.go
AlekSi: well, that wasn't correct :)

agents/mongodb/internal/profiler/aggregator/aggregator_test.go
```diff
@@ -94,4 +96,55 @@ func TestAggregator(t *testing.T) {
...
+		agentID := "test-agent"
+		startPeriod := time.Now()
+		aggregator := New(startPeriod, agentID, logrus.WithField("component", "test"))
+		aggregator.Start()
+		defer aggregator.Stop()
+		ctx := context.TODO()
+		err := aggregator.Add(ctx, proto.SystemProfile{
+			NscannedObjects: 2,
+			Nreturned:       3,
+			Ns:              "collection.people߿�\xff\\ud83d\xdd",
```
AlekSi, line 110: It is not possible to create that collection

agents/mysql/perfschema/perfschema.go
```diff
@@ -267,18 +266,16 @@ func makeBuckets(current, prev map[string]*eventsStatementsSummaryByDigest, l *l
...
 		}
 
-		fingerprint, truncated := truncate.Query(*currentESS.DigestText)
 		count := inc(currentESS.CountStar, prevESS.CountStar)
 		mb := &agentpb.MetricsBucket{
 			Common: &agentpb.MetricsBucket_Common{
 				Schema:                 pointer.GetString(currentESS.SchemaName), // TODO can it be NULL?
 				Queryid:                *currentESS.Digest,
-				Fingerprint:            fingerprint,
+				Fingerprint:            *currentESS.DigestText,
```
AlekSi: But we still need to truncate examples in getNewBuckets

BupycHuk:
So we have made mistake here https://github.com/percona/pmm-agent/commit/e25cd19c109d6ae9c98244354490058b7e9772d0#diff-a1abb864fee2d446721b2bc242151f6fR215 and we should not use isTruncated field in PostgreSQL, right?
Should we truncate fingerprints at all?

AlekSi:
PMM-5365 is about problems with examples, so let's focus on them. I don't think broken encoding is possible in fingerprints (column and table names, etc). For (maybe) truncating fingerprints we have https://jira.percona.com/browse/PMM-5283.

AlekSi:
> So we have made mistake here e25cd19#diff-a1abb864fee2d446721b2bc242151f6fR215

Looks like yes. But again, let's not change that now, and discuss that in the scope of PMM-5283. Maybe we will re-define IsTruncated. Or maybe @askomorokhov will remember why we have fingerprint as query example type. 😆

askomorokhov:
> > So we have made mistake here [e25cd19](https://github.com/percona/pmm-agent/commit/e25cd19c109d6ae9c98244354490058b7e9772d0)#diff-a1abb864fee2d446721b2bc242151f6fR215
> 
> Looks like yes. But again, let's not change that now, and discuss that in the scope of PMM-5283. Maybe we will re-define IsTruncated. Or maybe @askomorokhov will remember why we have fingerprint as query example type. 😆

It is for logging when the user allowed/disallowed collect examples.
https://github.com/percona/qan-api2/blob/master/migrations/sql/01_init.up.sql#L36

askomorokhov:
> So we have made mistake here [e25cd19#diff-a1abb864fee2d446721b2bc242151f6fR215](https://github.com/percona/pmm-agent/commit/e25cd19c109d6ae9c98244354490058b7e9772d0#diff-a1abb864fee2d446721b2bc242151f6fR215) and we should not use isTruncated field in PostgreSQL, right?
> Should we truncate fingerprints at all?

AFAIR, fingerprints in MySQL can be done from a truncated query example.

agents/mongodb/internal/profiler/aggregator/aggregator_test.go
```diff
@@ -94,4 +95,60 @@ func TestAggregator(t *testing.T) {
...
+		result := aggregator.createResult(ctx)
+
+		require.Equal(t, 1, len(result.Buckets))
+		assert.True(t, utf8.ValidString(result.Buckets[0].Common.Example))
+		assert.Equal(t, report.Result{
+			Buckets: []*agentpb.MetricsBucket{
+				{
+					Common: &agentpb.MetricsBucket_Common{
+						Queryid:             result.Buckets[0].Common.Queryid,
+						Fingerprint:         "FIND people name_\377",
```
AlekSi, line 124: I wonder what `\377` means there

agents/mysql/perfschema/perfschema_test.go
```diff
@@ -381,4 +389,50 @@ func TestPerfSchema(t *testing.T) {
 		expected.Common.Queryid = digests[expected.Common.Fingerprint]
 		tests.AssertBucketsEqual(t, expected, actual)
 	})
+
+	t.Run("Invalid UTF-8", func(t *testing.T) {
+		m := setup(t, db)
+
+		_, err := db.Exec("SELECT /* AllCities */ * FROM city where Name='BÃ¼rki'")
```
AlekSi, line 396: ⚠️ That's valid UTF-8 byte sequence: https://play.golang.org/p/iJsIPPkI2Ov I bet old code would pass that test too.

client/channel/channel_test.go
```diff
@@ -91,6 +93,54 @@ func setup(t *testing.T, connect func(agentpb.Agent_ConnectServer) error, expect
...
+				Example:     query,
+			},
+			Mysql: &agentpb.MetricsBucket_MySQL{},
+		},
+	}
+	resp := channel.SendRequest(rq)
+	assert.NotNil(t, resp)
+
+	rq.MetricsBucket = []*agentpb.MetricsBucket{
+		{
```
AlekSi, line 130: I think we used the other style before:
```
rq.MetricsBucket = []*agentpb.MetricsBucket{{
// ...
}}
```

agents/mongodb/internal/profiler/aggregator/aggregator_test.go
BupycHuk: our mongodb profiler has own escaper, I tried to find out where it's located, but couldn't find it.

AlekSi: Ok, let's not spend time on that now

agents/mysql/perfschema/perfschema_test.go
BupycHuk: Interesting thing here that channel is failing if we don't pass it through our truncate.Query function.

AlekSi: That sounds concerning. Let's _do_ spend time on investigating that :) Maybe create two different tests.

BupycHuk: checked it again and yeah, you are right, but for the some reason perfschema truncates query if there is some non-utf8 character. Trying to investigate it.

BupycHuk: So I'm going to remove this tests because it doesn't make sense.

AlekSi: I think we can keep this or similar (with \xff) test to track MySQL and MariaDB behaviour. It's just that it is not "Invalid UTF-8".

client/channel/channel_test.go
BupycHuk: fixed

agents/mysql/perfschema/perfschema_test.go
```diff
@@ -381,4 +383,61 @@ func TestPerfSchema(t *testing.T) {
...
+		assert.InDelta(t, 0, actual.Mysql.MLockTimeSum, 0.09)
+		expected := &agentpb.MetricsBucket{
+			Common: &agentpb.MetricsBucket_Common{
+				Fingerprint:            "SELECT * FROM `t1` WHERE `col1` = ?",
+				Schema:                 "world",
+				AgentId:                "agent_id",
+				PeriodStartUnixSecs:    1554116340,
+				PeriodLengthSecs:       60,
+				AgentType:              inventorypb.AgentType_QAN_MYSQL_PERFSCHEMA_AGENT,
+				Example:                "SELECT /* t1 */ * FROM t1 where col1='Bu", // Perf schema truncates queries with non-utf8 characters.
```
BupycHuk, line 418: @AlekSi this is how perfschema truncates query.

agents/mongodb/internal/profiler/aggregator/aggregator.go
AlekSi: @BupycHuk I think we should revert fingerprint truncation – and maybe even query truncation as it is should not be possible to have UTF-8 related problems in MongoDB – it uses _only_ UTF-8.

client/channel/channel_test.go
```diff
@@ -91,6 +93,52 @@ func setup(t *testing.T, connect func(agentpb.Agent_ConnectServer) error, expect
 	return
 }
 
+func TestAgentRequestWithTruncatedInvalidUTF8(t *testing.T) {
+	fingerprint, _ := truncate.Query("SELECT * FROM contacts t0 WHERE t0.person_id = '?';")
+	invalidQuery := "SELECT * FROM contacts t0 WHERE t0.person_id = '߿�\xff\\uD83D\xddÃ¼\xf1'"
```
AlekSi, line 98: ^

PR #172, opened 2020-06-18
Add comments to make it easier to use check-dev

AlekSi:


AlekSi:
Killed CI to stop useless builds

PR #179, opened 2020-08-14
PMM-6336 Fix Disable query example for Perf Schema.

askomorokhov:
https://jira.percona.com/browse/PMM-6336

FB: https://github.com/Percona-Lab/pmm-submodules/pull/1029

With query example:

Screen Shot 2020-08-13 at 2 01 06 PM

Without query example:

Screen Shot 2020-08-13 at 1 54 45 PM

### DIFF
--- a/agents/mysql/perfschema/perfschema.go
+++ b/agents/mysql/perfschema/perfschema.go
@@ -63,6 +63,15 @@ type Params struct {
 	DisableQueryExamples bool
 }
 
+// NewPerfSchemaParams holds all required parameters to instantiate a new PerfSchema
+type NewPerfSchemaParams struct {
+	Querier              *reform.Querier
+	DBCloser             io.Closer
+	AgentID              string
+	DisableQueryExamples bool
+	LogEntry             *logrus.Entry
+}
+
 const queryTag = "pmm-agent:perfschema"
 
 // New creates new PerfSchema QAN service.
@@ -77,16 +86,25 @@ func New(params *Params, l *logrus.Entry) (*PerfSchema, error) {
 	reformL := sqlmetrics.NewReform("mysql", params.AgentID, l.Tracef)
 	// TODO register reformL metrics https://jira.percona.com/browse/PMM-4087
 	q := reform.NewDB(sqlDB, mysql.Dialect, reformL).WithTag(queryTag)
-	return newPerfSchema(q, sqlDB, params.AgentID, params.DisableQueryExamples, l), nil
+
+	newParams := &NewPerfSchemaParams{
+		Querier:              q,
+		DBCloser:             sqlDB,
+		AgentID:              params.AgentID,
+		DisableQueryExamples: params.DisableQueryExamples,
+		LogEntry:             l,
+	}
+	return newPerfSchema(newParams), nil
+
 }
 
-func newPerfSchema(q *reform.Querier, dbCloser io.Closer, agentID string, disableQueryExamples bool, l *logrus.Entry) *PerfSchema {
+func newPerfSchema(params *NewPerfSchemaParams) *PerfSchema {
 	return &PerfSchema{
-		q:                    q,
-		dbCloser:             dbCloser,
-		agentID:              agentID,
-		disableQueryExamples: disableQueryExamples,
-		l:                    l,
+		q:                    params.Querier,
+		dbCloser:             params.DBCloser,
+		agentID:              params.AgentID,
+		disableQueryExamples: params.DisableQueryExamples,
+		l:                    params.LogEntry,
 		changes:              make(chan agents.Change, 10),
 		historyCache:         newHistoryCache(retainHistory),
 		summaryCache:         newSummaryCache(retainSummaries),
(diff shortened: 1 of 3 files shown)


agents/mysql/perfschema/perfschema_test.go
```diff
@@ -453,4 +453,19 @@ func TestPerfSchema(t *testing.T) {
 		require.NoError(t, err)
 		tests.LogTable(t, structs)
 	})
+
+	t.Run("DisableQueryExamples", func(t *testing.T) {
+		m := setup(t, db, true)
```
AlekSi, line 458: That's the problem with positional boolean arguments: does this `true` means disable or enable – and then disable or enable _what_ exactly? You can't tell without going to `setup` function implementation and then hoping to `newPerfSchema`. That's why positional boolean arguments are often considered a bad practice.

There are one and a half typical solutions in Go:
1. Use custom type and typed constants with good names. Unfortunately, due to [how underlying types and assignability works](https://golang.org/ref/spec#Assignability), that will not prevent one from using `true`/`false` directly: https://goplay.space/#hMnfuzpwZ-T. So that's only half of the solution.
2. Use params/options structures. As a bonus point, that often adds useful zero values.

We already have Params struct there, so let's use the same approach and introduce a new type (`newParams`?) that will contain all `newPerfSchema` parameters.

AlekSi:
Oh, and please fix CI

agents/mysql/perfschema/perfschema_test.go
```diff
@@ -453,4 +453,19 @@ func TestPerfSchema(t *testing.T) {
...
+		_, err = db.Exec("SELECT 1, 2, 3, 4, id FROM city WHERE id = 1")
+		require.NoError(t, err)
+
+		require.NoError(t, m.refreshHistoryCache())
+
+		buckets, err := m.getNewBuckets(time.Date(2019, 4, 1, 10, 59, 0, 0, time.UTC), 60)
+		require.NoError(t, err)
+
+		for _, b := range buckets {
+			assert.Empty(t, b.Common.Example)
```
AlekSi, line 468: I think it will not hurt to check that only example is empty, but something else is present

agents/mysql/perfschema/perfschema.go
```diff
@@ -63,6 +63,15 @@ type Params struct {
 	DisableQueryExamples bool
 }
 
+// NewPerfSchemaParams holds all required parameters to instantiate a new PerfSchema
+type NewPerfSchemaParams struct {
```
AlekSi, line 67: Please un-export it – symbols should not be exposed unless they have to

agents/mysql/perfschema/perfschema_test.go
AlekSi: ^

agents/mysql/perfschema/perfschema_test.go
```diff
@@ -33,6 +33,11 @@ import (
 	"github.com/percona/pmm-agent/utils/tests"
 )
 
+const (
+	disableQueryExamples = true
+	enableQueryExamples  = false
```
AlekSi, line 38: There is no need for those constants:
* `newPerfSchema(q, true)` is not clear;
* `newPerfSchema(&newPerfSchemaParams{Querier: q, DisableQueryExamples: true}` is clear;
* `newPerfSchema(&newPerfSchemaParams{Querier: q, DisableQueryExamples: disableQueryExamples}` is less clear because that's too much – it looks like `disableQueryExamples` is a variable that is set to true or false somewhere above :)

BupycHuk: So we use this variable to pass to setup function and we will have `setup(..., true)` or `setup(..., false)` if we remove this constants.

agents/mysql/perfschema/perfschema_test.go
BupycHuk: I see that @percona-csalguero already added a few lines above to check fingerprint. What else should be checked?

agents/mysql/perfschema/perfschema_test.go
AlekSi: What variable?

agents/mysql/perfschema/perfschema_test.go
AlekSi: I missed that :)

agents/mysql/perfschema/perfschema_test.go
BupycHuk: i meant constants.

AlekSi: So how can we avoid passing positional boolean arguments again? ;)

BupycHuk: new structure?

AlekSi: Or _typed_ constant. Or just a different `setup` method.

AlekSi:
CI fails

AlekSi (approved):

PR #181, opened 2020-08-19
PMM-6038 PT summary.

JiriCtvrtka:


actions/pt_summary_action.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	buf := new(bytes.Buffer)
+	cmd := exec.CommandContext(ctx, "pt-summary")
+	cmd.Dir = cfg.Paths.PTSummary
+	cmd.Stdout = buf
+	cmd.Stderr = new(bytes.Buffer)
+	if err := cmd.Run(); err != nil {
```
BupycHuk, line 62: we should pass
```
	pdeathsig.Set(cmd, unix.SIGKILL)
```

actions/pt_summary_action.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, err
```
BupycHuk, line 63: Should we return nil as result? Is it possible to have non-zero exit code but usefull output?

actions/pt_summary_action.go
```diff
@@ -0,0 +1,69 @@
...
+// Run runs an Action and returns output and error.
+func (a *ptSummaryAction) Run(ctx context.Context) ([]byte, error) {
+	l := logrus.WithField("component", "pt-summary")
+	cfg, _, err := config.Get(l)
+	if err != nil {
+		return nil, err
+	}
+
+	buf := new(bytes.Buffer)
+	cmd := exec.CommandContext(ctx, "pt-summary")
```
BupycHuk, line 58: Maybe it's better to extract logic related to starting, stopping and getting logs into separate structure? We will have other similar actions.

actions/pt_summary_action_test.go
```diff
@@ -0,0 +1,38 @@
...
+	t.Skip()
+
+	a := NewPTSummaryAction("")
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	res, err := a.Run(ctx)
+	require.NoError(t, err)
+	require.NotEmpty(t, res)
+	require.Regexp(t, regexp.MustCompile("# Percona Toolkit System Summary Report ######################"), string(res))
```
BupycHuk, line 37: We can use strings.contains instead of regexp.

config/config.go
```diff
@@ -89,7 +89,8 @@ type Paths struct {
 	ProxySQLExporter string `yaml:"proxysql_exporter"`
 	RDSExporter      string `yaml:"rds_exporter"`
 
-	TempDir string `yaml:"tempdir"`
+	TempDir   string `yaml:"tempdir"`
+	PTSummary string `yaml:"pt_summary"`
```
BupycHuk, line 93: let's keep it in a separate block

BupycHuk (changes requested):

config/config_test.go
```diff
@@ -84,6 +84,7 @@ func TestGet(t *testing.T) {
 		actual, configFilepath, err := get([]string{
 			"--id=agent-id",
 			"--server-address=127.0.0.1",
+			"--paths-pt-summary=/usr/local/percona/pmm2/pt-summary",
```
BupycHuk, line 87: let's remove it and check default value. To learn where it placed ask @EvgeniyPatlan

config/config_test.go
```diff
@@ -118,6 +120,9 @@ func TestGet(t *testing.T) {
 			Server: Server{
 				Address: "127.0.0.1",
 			},
+			Paths: Paths{
```
BupycHuk, line 123: Let's remove it too, we should have default value

BupycHuk (changes requested):

client/client.go
```diff
@@ -300,6 +300,9 @@ func (c *Client) processChannelRequests() {
 			case *agentpb.StartActionRequest_MongodbQueryGetcmdlineoptsParams:
 				action = actions.NewMongoDBQueryAdmincommandAction(p.ActionId, params.MongodbQueryGetcmdlineoptsParams.Dsn, "getCmdLineOpts", 1)
 
+			case *agentpb.StartActionRequest_PtSummaryParams:
+				action = actions.NewProcessAction(p.ActionId, c.cfg.Paths.PTSummary, []string{})
```
BupycHuk, line 304:
```suggestion
				action = actions.NewProcessAction(p.ActionId, c.cfg.Paths.PTSummary, new([]string))
```

BupycHuk (approved):
Only one minor comment

client/client.go
AlekSi: That will not compile

JiriCtvrtka: Yes, I already discard it. Thanks.

JiriCtvrtka:
FB: https://github.com/Percona-Lab/pmm-submodules/pull/1065

PR #183, opened 2020-09-01
PMM-6460 Pg stat monitor

JiriCtvrtka:


agents/postgres/pgstatmonitor/models.go
```diff
@@ -0,0 +1,90 @@
...
+	TempBlksRead      int64   `reform:"temp_blks_read"`
+	TempBlksWritten   int64   `reform:"temp_blks_written"`
+	BlkReadTime       float64 `reform:"blk_read_time"`
+	BlkWriteTime      float64 `reform:"blk_write_time"`
+	Host              string  `reform:"host"`
+	ClientIP          string  `reform:"client_ip"`
+	RespCalls         string  `reform:"resp_calls"`
+	TablesNames       string  `reform:"tables_names"`
+	WaitEvent         string  `reform:"wait_event"`
+	PgStatAggDatabase string  `reform:"pg_stat_agg_database"`
```
BupycHuk, line 71: AFAIK these fields are not a part of pg_stat_monitor view.

agents/postgres/pgstatmonitor/models.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// pgStatMonitorExtended contains pgStatMonitor data and extends it with database, username and tables data.
+// It's made for performance reason.
+type pgStatMonitorExtended struct {
+	pgStatMonitor
+
+	Database         string
+	Username         string
+	Tables           []string
```
BupycHuk, line 83: We don't need this field here since we already have it in pgStatMonitor

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		count := float32(currentPSS.Calls - prevPSS.Calls)
+		switch {
+		case count == 0:
+			// Another way how this is possible is if pg_stat_monitor was truncated,
+			// and then the same number of queries were made.
+			// Currently, we can't differentiate between those situations.
+			l.Tracef("Skipped due to the same number of queries: %s.", currentPSS)
+			continue
+		case count < 0:
```
BupycHuk, line 199: pg stat monitor gives us one more option to detect truncate. Bucket started time.

agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -0,0 +1,257 @@
...
+
+			// use previous values
+			c.Tables = p.Tables
+			c.Query, c.IsQueryTruncated = p.Query, p.IsQueryTruncated
+		} else {
+			newN++
+
+			// do not extract tables again if we saw this query during this iteration already
+			if tables[c.QueryID] == nil {
+				tables[c.QueryID] = extractTables(c.Query, ssc.l)
```
BupycHuk, line 134: we don't need to use extractTables in pg stat monitor since we already have them in the view.

BupycHuk (changes requested):

agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -0,0 +1,232 @@
...
+			newN++
+
+			// do not extract tables again if we saw this query during this iteration already
+			if tables[c.QueryID] == nil {
+				tables[c.QueryID] = c.TablesNames
+			} else {
+				newSharedN++
+			}
+
+			c.TablesNames = tables[c.QueryID]
```
BupycHuk, line 138: No need to override it, since we get this value from pg_stat_monitor.

utils/tests/postgresql.go
```diff
@@ -84,6 +84,21 @@ func PostgreSQLVersion(tb testing.TB, db *sql.DB) string {
...
+
+	var version string
+	err := db.QueryRow("SELECT /* pmm-agent-tests:PostgreSQLVersion */ version()").Scan(&version)
+	require.NoError(tb, err)
+
+	if strings.Contains(version, "Red Hat") {
+		return "percona"
+	}
+
+	return "postgre"
```
BupycHuk, line 99:
```suggestion
	return "PostgreSQL"
```

utils/tests/postgresql.go
```diff
@@ -84,6 +84,21 @@ func PostgreSQLVersion(tb testing.TB, db *sql.DB) string {
...
 
+// PostgreSQLVendor returns PostgreSQL vendor detected by first string from version.
+func PostgreSQLVendor(tb testing.TB, db *sql.DB) string {
+	tb.Helper()
+
+	var version string
+	err := db.QueryRow("SELECT /* pmm-agent-tests:PostgreSQLVersion */ version()").Scan(&version)
+	require.NoError(tb, err)
+
+	if strings.Contains(version, "Red Hat") {
```
BupycHuk, line 95: Let's add a comment that there is no way to detect other way.

BupycHuk (changes requested):

agents/postgres/pgstatstatements/pgstatstatements_test.go
```diff
@@ -133,6 +133,18 @@ func TestPGStatStatementsQAN(t *testing.T) {
 		}
 	}
 
+	// Need detect vendor because result for mSharedBlksReadSum are diffrent for diffrent images for postgres.
+	engineVendor := tests.PostgreSQLVendor(t, sqlDB)
+	assert.NotEmpty(t, engineVendor)
+
+	var mSharedBlksReadSum float32
```
BupycHuk, line 140:
```suggestion
	var mSharedBlksHitSum float32
```

agents/postgres/pgstatmonitor/stat_monitor_cache.go
BupycHuk: @JiriCtvrtka ^

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		count := float32(currentPSS.Calls - prevPSS.Calls)
+		switch {
+		case count == 0:
+			// Another way how this is possible is if pg_stat_monitor was truncated,
+			// and then the same number of queries were made.
+			// Currently, we can't differentiate between those situations.
+			l.Tracef("Skipped due to the same number of queries: %s.", currentPSS)
+			continue
+		case count < 0 || (time.Now().Sub(currentPSS.BucketStartTime).Minutes() < 1):
```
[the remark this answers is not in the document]
JiriCtvrtka: Changed to time.Now().Sub(currentPSS.BucketStartTime).Seconds() > 60 and because of 1 minute interval for query pg stat monitor and test passed. Hope it is fine.

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,302 @@
...
+			l.Debugf("Truncate detected. Treating as a new query: %s.", currentPSS)
+			prevPSS = new(pgStatMonitorExtended)
+			count = float32(currentPSS.Calls)
+		case prevPSS.Calls == 0:
+			l.Debugf("New query: %s.", currentPSS)
+		default:
+			l.Debugf("Normal query: %s.", currentPSS)
+		}
+
+		fingerprint := query.Fingerprint(currentPSS.Query)
```
askomorokhov, line 228: Make a fingerprint is a quite heavy operation. I'd swap it in the next way:
```
fingerprint := currentPSS.Query
example := ""
if !pgsmNormalizedQuery {
 			fingerprint = query.Fingerprint(currentPSS.Query)
 			example = currentPSS.Query
}
```

askomorokhov (changes requested):

agents/postgres/pgstatmonitor/pgstatmonitor.go
JiriCtvrtka: Yes true, changed. Thanks.

askomorokhov (approved):
Looks good to me.

.travis.yml
```diff
@@ -41,6 +41,8 @@ env:
     - MYSQL_IMAGE=mysql:5.6 MONGO_IMAGE=mongo:3.6 POSTGRES_IMAGE=postgres:10 PMM_SERVER_IMAGE=percona/pmm-server:2.0.0
     - MYSQL_IMAGE=mysql:5.7 MONGO_IMAGE=mongo:4.0 POSTGRES_IMAGE=postgres:11 PMM_SERVER_IMAGE=percona/pmm-server:2.0.1
     - MYSQL_IMAGE=mysql:8.0 MONGO_IMAGE=mongo:4.2 POSTGRES_IMAGE=postgres:12 PMM_SERVER_IMAGE=percona/pmm-server:2.1.0
+    - MYSQL_IMAGE=mysql:5.7 MONGO_IMAGE=mongo:4.0 POSTGRES_IMAGE=perconalab/percona-distribution-postgresql:11 PG_PRELOADED_LIBS=pg_stat_statements,pg_stat_monitor PMM_SERVER_IMAGE=percona/pmm-server:2.0.1
+    - MYSQL_IMAGE=mysql:8.0 MONGO_IMAGE=mongo:4.2 POSTGRES_IMAGE=perconalab/percona-distribution-postgresql:12 PG_PRELOADED_LIBS=pg_stat_statements,pg_stat_monitor PMM_SERVER_IMAGE=percona/pmm-server:2.1.0
```
AlekSi, line 45: Is it "latest upstreams" as the comment above says, or that's "Percona" as comment below says?

agents/postgres/pgstatmonitor/models.go
```diff
@@ -0,0 +1,99 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pgstatmonitor
+
+import (
+	"fmt"
+	"time"
+
+	pq_types "github.com/mc2soft/pq-types"
```
AlekSi, line 22: We should not use that library directly. If we don't have required code somewhere already (please search), let's just copy the part we need

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+		count := float32(currentPSS.Calls - prevPSS.Calls)
+		switch {
+		case count == 0:
+			// Another way how this is possible is if pg_stat_monitor was truncated,
+			// and then the same number of queries were made.
+			// Currently, we can't differentiate between those situations.
+			l.Tracef("Skipped due to the same number of queries: %s.", currentPSS)
+			continue
+		case count < 0 || (time.Now().Sub(currentPSS.BucketStartTime).Seconds() > 60):
```
AlekSi, line 218: Why we need that condition there, but not for pg_stat_statements?

.travis.yml
JiriCtvrtka: Yes, it was confusing. I added new group and left comment.

agents/postgres/pgstatmonitor/models.go
JiriCtvrtka: I searched and found that StringArray is also contained in pg package. So now I using it from here.

agents/postgres/pgstatmonitor/pgstatmonitor.go
JiriCtvrtka: I am not sure why it is not in pg_stat_statements too, but Nurlan highlight that for this can there should be also condition for start bucket time. I checked interval and add this condition with more than 60 seconds to create new bucket.

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,302 @@
...
+		case prevPSS.Calls == 0:
+			l.Debugf("New query: %s.", currentPSS)
+		default:
+			l.Debugf("Normal query: %s.", currentPSS)
+		}
+
+		fingerprint := currentPSS.Query
+		example := ""
+		if !pgsmNormalizedQuery {
+			fingerprint = query.Fingerprint(currentPSS.Query)
```
BupycHuk, line 231: at this moment queries already will be truncated.
+ Let's cache fingerprint for the same queries. Fot this let's add additional field in `PgStatMonitorExtended` `fingerpint` and let's move these lines to https://github.com/percona/pmm-agent/pull/183/files#diff-c7bb32cfdcfd2dcb1693bf29def551dcR124 and use values from previous bucket here. https://github.com/percona/pmm-agent/pull/183/files#diff-c7bb32cfdcfd2dcb1693bf29def551dcR120.

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,302 @@
...
+				Username:    currentPSS.Username,
+				Queryid:     currentPSS.QueryID,
+				NumQueries:  count,
+				AgentType:   inventorypb.AgentType_QAN_POSTGRESQL_PGSTATMONITOR_AGENT,
+			},
+			Postgresql: new(agentpb.MetricsBucket_PostgreSQL),
+		}
+
+		if example != "" && !disableQueryExamples {
+			example, truncated := truncate.Query(example)
```
BupycHuk, line 249: let's truncate it here https://github.com/percona/pmm-agent/pull/183/files#diff-c7bb32cfdcfd2dcb1693bf29def551dcR124 and use `IsQueryTruncated` from `pgStatMonitorExtended`

BupycHuk (changes requested):

.travis.yml
AlekSi: Exactly. We did not have Percona PostgreSQL before, but we have it now, so it should be in the block below

BupycHuk (approved):

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,289 @@
...
+
+		buckets[i] = b
+	}
+
+	return buckets, nil
+}
+
+// makeBuckets uses current state of pg_stat_monitor table and accumulated previous state
+// to make metrics buckets.
+//
```
askomorokhov, line 197: Looks like an empty comment string

askomorokhov (approved):
Looks good to me.

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,288 @@
...
+	return buckets, nil
+}
+
+// makeBuckets uses current state of pg_stat_monitor table and accumulated previous state
+// to make metrics buckets.
+func (m *PGStatMonitorQAN) makeBuckets(current, prev map[string]*pgStatMonitorExtended) []*agentpb.MetricsBucket {
+	res := make([]*agentpb.MetricsBucket, 0, len(current))
+	now := time.Now()
+
+	for queryID, currentPSS := range current {
```
AlekSi, line 201: PSS stands for Pg Stat Statement. It should be PSM for all pg_stat_monitor code

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,288 @@
...
+			m.l.Infof("Context canceled.")
+			return
+
+		case <-t.C:
+			if !running {
+				m.changes <- agents.Change{Status: inventorypb.AgentStatus_STARTING}
+			}
+
+			lengthS := uint32(math.Round(wait.Seconds())) // round 59.9s/60.1s to 60s
+			buckets, err := m.getNewBuckets(ctx, start, lengthS)
```
AlekSi, line 144: TBH, it does not look like we align buckets correctly. Why we start bucket at `:00` seconds when we have bucket start time, for example?

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+		count := float32(currentPSS.Calls - prevPSS.Calls)
+		switch {
+		case count == 0:
+			// Another way how this is possible is if pg_stat_monitor was truncated,
+			// and then the same number of queries were made.
+			// Currently, we can't differentiate between those situations.
+			m.l.Tracef("Skipped due to the same number of queries: %s.", currentPSS)
+			continue
+		case count < 0 || (now.Sub(currentPSS.BucketStartTime) > queryStatMonitor):
```
AlekSi, line 214: It looks like that second half of the condition is too aggressive – it assumes that all code below runs instantly, isn't it?

Please move the second half to a separate case with a different logging line (but the same logic) and test with debug logging

BupycHuk: What do you mean by saying that all code below runs instantly?

AlekSi: s/below/above. We compare values down to nanoseconds, and seems concerning… So let's do the second half of my comment above to see if that's a problem in practice

agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -0,0 +1,237 @@
...
+	"time"
+
+	"github.com/percona/pmm-agent/utils/truncate"
+
+	"github.com/AlekSi/pointer"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"gopkg.in/reform.v1"
+
+	"github.com/percona/go-mysql/query"
```
BupycHuk, line 31: How we missed that?

JiriCtvrtka: It is called from fingerprint. Postgres package doesnt had fingerprint method.

https://github.com/percona/pmm-agent/pull/183/files#diff-c7bb32cfdcfd2dcb1693bf29def551dcR132

Or do you mean something else?

BupycHuk: we use mysql package in postgresql. It should not work this way.

JiriCtvrtka: It was discussed with @askomorokhov, because he help me with fingerprints. So please @BupycHuk, what you suggest how to do that? Without this package.

BupycHuk: Use https://github.com/lfittl/pg_query_go to normalize query. PTAL https://github.com/lfittl/pg_query_go/blob/master/normalize_test.go#L23

askomorokhov: 🤦  Yeah, it was a suggestion to it temporary - just to run.
Definitely we need to use https://github.com/lfittl/pg_query_go here for prod.

JiriCtvrtka: @askomorokhov Sorry I missed that it is temporarily.

PR #203, opened 2020-10-23
PMM-4172 PT MySQL summary.

yareach:


actions/mysql_show_table_status_action_test.go
```diff
@@ -82,6 +82,19 @@ func TestShowTableStatus(t *testing.T) {
 				nil, "", "",
 			}, actual[1])
 
+		case "10.2":
```
BupycHuk, line 85: this case can be added to the case below.

BupycHuk (changes requested):
Looks good to me, only one change request.

actions/mysql_show_table_status_action_test.go
yareach: They are different.

client/client.go
```diff
@@ -304,6 +306,27 @@ func (c *Client) processChannelRequests() {
...
+					addr := cfg.Addr
+
+					// Strips the port info (if ':' found)
+					if i := strings.Index(addr, ":"); i > -1 {
+						addr = addr[:i]
+					}
+
+					// Action with path and --host, --user, --password data
+					action = actions.NewProcessAction(p.ActionId, c.cfg.Paths.PTMySqlSummary,
+						[]string{"--host=" + addr, "--user=" + cfg.User, "--password=" + cfg.Passwd})
```
percona-csalguero, line 325: If MySQL is not running on the standard port, will this work sin port is missing here?

percona-csalguero (changes requested):

client/client.go
```diff
@@ -304,6 +306,27 @@ func (c *Client) processChannelRequests() {
 			case *agentpb.StartActionRequest_PtSummaryParams:
 				action = actions.NewProcessAction(p.ActionId, c.cfg.Paths.PTSummary, []string{})
 
+			case *agentpb.StartActionRequest_PtMysqlSummaryParams:
+				// Parses the DSN string
+				cfg, err := mysql.ParseDSN(params.PtMysqlSummaryParams.Dsn)
```
BupycHuk, line 311: Why do we pass dsn and parse it here?

BupycHuk: https://github.com/percona/pmm/pull/600#discussion_r518741604

BupycHuk, line 311: Can we extract all this logic somewhere and write tests for it.
we also don't support sockets here.

BupycHuk (changes requested):

client/client.go
yareach: Added the port to the arguments.

BupycHuk:
@yareach I requested changes 7 minutes ago, there are no changes

client/client.go
yareach: Are you asking my to write a function to handle the parsing of this case?

BupycHuk: Yes, generating flags from input parameters.

client/client.go
```diff
@@ -538,3 +542,32 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
...
+
+	if pParams.Address != "" {
+		args = append(args, "--host="+pParams.Address)
+	}
+
+	if pParams.Port > 0 && pParams.Port <= 65535 {
+		args = append(args, "--port="+strconv.FormatUint(uint64(pParams.Port), 10))
+	}
+
+	if pParams.Socket != "" {
```
percona-csalguero, line 560: Can we use socket AND host and port? 
Shouldn't we check and use only one of them?

percona-csalguero (changes requested):

client/client.go
yareach: I believe that if the one of them is not used, the string is empty or 0 in case of the port.
If there is another pattern, I can change it but I'd need clear specification to be able to create right conditions. Can you provide me with such information?

percona-csalguero: I would check that if host and port are valid and also we have socket, just use socket and don't use host+port

yareach: If socket is valid => use just socket
If socket is not valid => check user and password and use them without socket.
Correct?

percona-csalguero: :+1:

BupycHuk (changes requested):

BupycHuk (approved):

percona-csalguero (approved):

client/client.go
```diff
@@ -538,3 +542,32 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
 var (
 	_ prometheus.Collector = (*Client)(nil)
 )
+
+// argListFromMySqlParams creates an array of strings from the pointer to the parameters for pt-mysql-sumamry
+func argListFromMySqlParams(pParams *agentpb.StartActionRequest_PTMySQLSummaryParams) []string {
+	var args []string
+
+	// Only adds the arguments are valid
+	// If socket valid, socket will be used. Otherwise it will try user and password.
```
AlekSi, line 551: Ok, but why? Socket can't be used with address (you either use TCP connection with address, or UNIX socket connection), but I think it can be used with username and password.

/cc @percona-csalguero

client/client.go
```diff
@@ -538,3 +542,32 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
 var (
 	_ prometheus.Collector = (*Client)(nil)
 )
+
+// argListFromMySqlParams creates an array of strings from the pointer to the parameters for pt-mysql-sumamry
+func argListFromMySqlParams(pParams *agentpb.StartActionRequest_PTMySQLSummaryParams) []string {
```
AlekSi, line 547: We typically don't add normal code after interface checking assertions like `_ prometheus.Collector = (*Client)(nil)` above. Please move it up

client/client.go
```diff
@@ -304,6 +305,9 @@ func (c *Client) processChannelRequests() {
 			case *agentpb.StartActionRequest_PtSummaryParams:
 				action = actions.NewProcessAction(p.ActionId, c.cfg.Paths.PTSummary, []string{})
 
+			case *agentpb.StartActionRequest_PtMysqlSummaryParams:
+				// Action with path and arguments list to run pt-mysql-summary
```
AlekSi, line 309: There are no comments for any other action above. Why add it there?

actions/mysql_show_table_status_action_test.go
```diff
@@ -50,11 +50,20 @@ func TestShowTableStatus(t *testing.T) {
...
 
 		var actual [][]interface{}
+
+		// Unmarshals the stream of bytes. Uses the following types:
+		//	bool, for JSON booleans
+		//	float64, for JSON numbers
+		//	string, for JSON strings
+		//	[]interface{}, for JSON arrays
+		//	map[string]interface{}, for JSON objects
+		//	nil for JSON null
```
AlekSi, line 60: Do we really have to quote documentation there?

client/client.go
```diff
@@ -538,3 +542,32 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
...
+	var args []string
+
+	// Only adds the arguments are valid
+	// If socket valid, socket will be used. Otherwise it will try user and password.
+	if pParams.Socket != "" {
+		args = append(args, "--socket="+pParams.Socket)
+	} else {
+		if pParams.Username != "" {
+			args = append(args, "--user="+pParams.Username)
+		}
```
percona-csalguero, line 557: Socket or Host + port.
User and pass can be used with socket and host.

client/client.go
```diff
@@ -538,3 +542,32 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
...
+
+		if pParams.Password != "" {
+			args = append(args, "--password="+pParams.Password)
```
percona-csalguero, line 560: Should we escape passwords? What if the password has a `"`

percona-csalguero (changes requested):

actions/mysql_show_table_status_action_test.go
```diff
@@ -50,11 +50,20 @@ func TestShowTableStatus(t *testing.T) {
...
 		err = json.Unmarshal(b, &actual)
 		require.NoError(t, err)
 		require.Len(t, actual, 2)
 
 		const createTime = "2019-06-10 12:04:29"
+		// The numbers used in the test vectors are with the decimal points deu to JSON comparison.
```
JiriCtvrtka, line 66: Little typo. DEU

client/client.go
yareach: Can you specify what exactly you would like to do there?

client/client.go
yareach: This requirement is not clear. OR or XOR? And brackets would be helpful too...

actions/mysql_show_table_status_action_test.go
yareach: If we have to... I don't know. But for sure we can.

client/client.go
yareach: I'm gonna fix this. I didn't know.

client/client.go
JiriCtvrtka: You using " and Carlos mention that password can also contain ". That is problem.

yareach: I'm not sure if I understand it well here. "" stands for an empty string. If the password contains ", it will be handled as a string. It may generate an error when passed to pt-mysql-summary. If it does, it will finish with an error.
My question is: What do you desire to do here? Replace, remove and/or generate an error code without execution?

percona-csalguero: A MySQL password can contain quotes and special chars. If we use just a `+` to concatenate the password, it might not work. 
Here is an example: https://play.golang.org/p/qlPjVEltESM

AlekSi: Carlos, will pt-mysql-summary accept `--password mycoolpass`, without `=`? If yes, we should append two strings to args: `args = append(args, "--password", pParams.Password)`. os/exec.Cmd will handle the rest

percona-csalguero: :+1: Yes, it would. @yareach  could you make it work as @AlekSi said?
Thanks.

client/client.go
mvsantos: pt-mysql-summary accepts any combination of socket/host/port/user/pass https://github.com/percona/percona-toolkit/blob/3.x/bin/pt-mysql-summary#L482-L503

client/client.go
yareach: Roger that.

actions/mysql_show_table_status_action_test.go
AlekSi: Please remove it: we don't quote documentation for library packages – it is assumed that Go developers know packages they use work, especially standard library packages

AlekSi (approved):
LGTM, but please remove extra comment

percona-csalguero (approved):

BupycHuk (approved):

percona-csalguero (approved):

askomorokhov (approved):

askomorokhov (approved):

actions/mysql_summary_action_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMySQLSummaryActionRun(t *testing.T) {
```
BupycHuk, line 27:
```suggestion
func TestPTMySQLSummaryActionRun(t *testing.T) {
```

actions/mysql_summary_action.go
```diff
@@ -0,0 +1,66 @@
...
+	id      string
+	command string
+	arg     []string
+}
+
+// NewMySQLAction creates a new process Action.
+//
+// MySQL Action, it's an abstract Action that can run an external commands.
+// This commands can be a shell script, script written on interpreted language, or binary file.
+func NewMySQLAction(id string, cmd string, arg []string) Action {
```
BupycHuk, line 38:
```suggestion
func NewPTMySQLAction(id string, cmd string, arg []string) Action {
```

client/client.go
```diff
@@ -561,6 +564,31 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
 	}
 }
 
+// argListFromMySqlParams creates an array of strings from the pointer to the parameters for pt-mysql-sumamry
+func argListFromMySqlParams(pParams *agentpb.StartActionRequest_PTMySQLSummaryParams) []string {
```
BupycHuk, line 568: let's move this into `PTMySQLAction`

BupycHuk (changes requested):

client/client_test.go
```diff
@@ -249,3 +249,38 @@ func TestArgListFromMongoDBParams(t *testing.T) {
 		})
 	}
 }
+
+func Test_argListFromMySqlParams(t *testing.T) {
```
BupycHuk, line 253: let's move this too

BupycHuk: and let's make it camelCase

JiriCtvrtka: :facepalm: yes sure, missed that.

actions/mysql_summary_action.go
```diff
@@ -0,0 +1,93 @@
...
+	cmd.Dir = "/"
+	pdeathsig.Set(cmd, unix.SIGKILL)
+
+	return cmd.CombinedOutput()
+}
+
+func (*mysqlSummaryAction) sealed() {}
+
+// argListFromMySqlParams creates an array of strings from the pointer to the parameters for pt-mysql-sumamry
+func argListFromMySqlParams(params *agentpb.StartActionRequest_PTMySQLSummaryParams) []string {
```
BupycHuk, line 71: We can make it method and don't pass any argument.

JiriCtvrtka: Ok will change it.

actions/mysql_summary_action.go
```diff
@@ -0,0 +1,97 @@
...
+	"os"
+	"os/exec"
+	"strconv"
+
+	"github.com/percona/pmm/api/agentpb"
+	"github.com/percona/pmm/utils/pdeathsig"
+	"golang.org/x/sys/unix"
+)
+
+type mysqlSummaryAction struct {
```
BupycHuk, line 30:
```suggestion
type ptMySQLSummaryAction struct {
```
please check everywhere

BupycHuk: rename the file

JiriCtvrtka: @BupycHuk pt_mysql_summary_action.go and pt_mysql_summary_action_test.go?

JiriCtvrtka: Renamed structure and files too.

BupycHuk (approved):

PR #209, opened 2020-11-13
PMM-5784 pt-pg-summary.

yareach:


percona-csalguero (approved):

client/client.go
```diff
@@ -305,6 +306,13 @@ func (c *Client) processChannelRequests() {
 			case *agentpb.StartActionRequest_PtSummaryParams:
 				action = actions.NewProcessAction(p.ActionId, c.cfg.Paths.PTSummary, []string{})
 
+			case *agentpb.StartActionRequest_PtMysqlSummaryParams:
```
BupycHuk, line 309: please update dependencies.

BupycHuk: and remove this line from this PR

client/client.go
```diff
@@ -535,6 +539,31 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
...
+	var args []string
+
+	// Only adds the arguments are valid
+
+	if pParams.Host != "" {
+		args = append(args, "--host", pParams.Host)
+	}
+
+	if pParams.Port > 0 && pParams.Port <= 65535 {
+		args = append(args, "--port", strconv.FormatUint(uint64(pParams.Port), 10))
```
percona-csalguero, line 553: Wasn't fmt.Sprintf shorter?

client/client_test.go
```diff
@@ -215,3 +215,35 @@ func TestGetActionTimeout(t *testing.T) {
...
+			Password: ""}, []string{"--host", "10.20.30.40", "--port", "555", "--username", "person"}},
+		{&agentpb.StartActionRequest_PTPgSummaryParams{Host: "10.20.30.40", Port: 555, Username: "",
+			Password: "secret"}, []string{"--host", "10.20.30.40", "--port", "555", "--password", "secret"}},
+		{&agentpb.StartActionRequest_PTPgSummaryParams{Host: "10.20.30.40", Port: 65536, Username: "",
+			Password: "secret"}, []string{"--host", "10.20.30.40", "--password", "secret"}},
+		{&agentpb.StartActionRequest_PTPgSummaryParams{Host: "", Port: 555, Username: "", Password: "secret"},
+			[]string{"--port", "555", "--password", "secret"}},
+
+		{&agentpb.StartActionRequest_PTPgSummaryParams{Host: "", Port: 0, Username: "", Password: ""}, []string{}},
+		{&agentpb.StartActionRequest_PTPgSummaryParams{Host: "", Port: 0, Username: "王华", Password: `"`},
```
percona-csalguero, line 238: Can we add a test with a password having blank spaces?
I think in that case the password is not being quoted.

percona-csalguero (changes requested):

client/client.go
yareach: In my opinion sprintf has more overhead because it has to parse the string first. This is just cast to 64 bits and then direct conversion to string.
Do you want me to change it?

BupycHuk: let's use `strconv.Itoa` fot better readability.

client/client_test.go
```diff
@@ -215,3 +215,37 @@ func TestGetActionTimeout(t *testing.T) {
...
+			[]string{"--username", "王华", "--password", `"`}},
+		{&agentpb.StartActionRequest_PTPgSummaryParams{Host: "10.20.30.40", Port: 555, Username: "",
+			Password: "   "}, []string{"--host", "10.20.30.40", "--port", "555", "--password", "   "}},
```
BupycHuk, line 241: it should not work this way, if we will have empty password we should pass flag.

BupycHuk (changes requested):

client/client_test.go
yareach: What do you want to do here? If there is a whitespace 0x20 within the password, don't pass the password? Or strip the spaces from left and right and if it becomes "", not to pass the password?

percona-csalguero (approved):

client/client_test.go
BupycHuk: yes, let's trim spaces from left and right and it becomes empty string do not pass the password.

BupycHuk (approved):

client/client.go
```diff
@@ -556,6 +560,31 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
 	}
 }
 
+// argListFromPgParams creates an array of strings from the pointer to the parameters for pt-pg-sumamry
+func argListFromPgParams(pParams *agentpb.StartActionRequest_PTPgSummaryParams) []string {
+	var args []string
+
+	// Only adds the arguments are valid
```
Dasio, line 567: For me it seems like useless comment.

client/client_test.go
Dasio: What about this?

Dasio (approved):

askomorokhov (approved):

BupycHuk (approved):

PR #210, opened 2020-11-25
PMM-5364 SSL support for Mongo.

JiriCtvrtka:


askomorokhov (changes requested):

BupycHuk (approved):

utils/templates/template.go
```diff
@@ -0,0 +1,118 @@
...
+		if err != nil {
+			return "", err
+		}
+
+		b, err := tr.RenderTemplate("dsn", dsn, templateParams)
+		if err != nil {
+			return "", err
+		}
+		dsn = string(b)
+		fmt.Println(dsn, templateParams)
```
percona-csalguero, line 115: Why we have this? It is a log?

percona-csalguero (changes requested):

utils/templates/template.go
BupycHuk: Removed

AlekSi (changes requested):
Tests fail on CI with different versions of MongoDB

artemgavrilov (approved):

Makefile
```diff
@@ -57,18 +57,19 @@ install-race:                   ## Install pmm-agent binary with race detector.
 
 TEST_PACKAGES ?= ./...
 TEST_FLAGS ?= -timeout=60s
+TEST_ENV_VARS ?= GODEBUG=x509ignoreCN=0
```
AlekSi, line 60: That line makes all testing efforts meaningless – essentially, we are not testing with certificates that our users could use.

Please generate valid certificate with https://github.com/FiloSottile/mkcert (`mkcert -install; mkcert -cert-file=XXX -key-file=XXX localhost 127.0.0.1`) in Makefile and run that target on CI.

AlekSi (changes requested):

Makefile
BupycHuk: regenerated certificates with mkcert

AlekSi: `mkcert -install` should be called on CI, and certificates should be generated there too, not committed in the repo – it will not work otherwise. Please take a moment to read mkcert's README

AlekSi (changes requested):
CI will fail

percona-csalguero (approved):

PR #212, opened 2020-12-08
PMM-4306 Pt mongodb summary

yareach:


client/client_test.go
```diff
@@ -215,3 +215,37 @@ func TestGetActionTimeout(t *testing.T) {
...
+		{&agentpb.StartActionRequest_PTMongoDBSummaryParams{Host: "10.20.30.40", Port: 65536, Username: "",
+			Password: "secret"}, []string{"-psecret", "10.20.30.40"}},
+		{&agentpb.StartActionRequest_PTMongoDBSummaryParams{Host: "", Port: 555, Username: "", Password: "secret"},
+			[]string{"-psecret"}},
+
+		{&agentpb.StartActionRequest_PTMongoDBSummaryParams{Host: "", Port: 0, Username: "", Password: ""}, []string{}},
+		{&agentpb.StartActionRequest_PTMongoDBSummaryParams{Host: "", Port: 0, Username: "王华", Password: `"`},
+			[]string{"--username", "王华", `-p"`}},
+		{&agentpb.StartActionRequest_PTMongoDBSummaryParams{Host: "10.20.30.40", Port: 555, Username: "person",
+			Password: "   "}, []string{"--username", "person", "-p   ", "10.20.30.40:555"}},
```
percona-csalguero, line 241: Sorry for being picky, can we test it with password = "has a blank" to check it is quoted ?

percona-csalguero (changes requested):

client/client.go
```diff
@@ -535,6 +540,35 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
...
+	var args []string
+
+	// Only adds the arguments are valid
+
+	if pParams.Username != "" {
+		args = append(args, "--username", pParams.Username)
+	}
+
+	if pParams.Password != "" {
+		// TBD change this line when pt-mongodb-summary is updated
```
BupycHuk, line 554:
```suggestion
		// TODO change this line when pt-mongodb-summary is updated
```

BupycHuk: How it should be changed? can we use `--password` here?

client/client_test.go
```diff
@@ -215,3 +215,37 @@ func TestGetActionTimeout(t *testing.T) {
 		})
 	}
 }
+
+func Test_argListFromMongoDBParams(t *testing.T) {
```
BupycHuk, line 219:
```suggestion
func TestArgListFromMongoDBParams(t *testing.T) {
```

client/client_test.go
BupycHuk, line 241: won't it fail calling pt-mongodb-summary?

BupycHuk (changes requested):

client/client.go
yareach: I mean, we would use it the same way as in for mySQL or postgreSQL, it means two strings to the array: { "--pasword", .Password }. The other PTs accept it without "=", the only exception is for pt-mongodb-summary's password param.
Yes, we can use --password but it will require "=" with this version of pt.

client/client_test.go
yareach: It will not quote it. If we put there %q, the password does not work. It works if we quote it from the command line but it does not, if we pass it from pmm-agent.
In my opinion, the command line has to parse it and identify the params. If you pass it from one app to another, it is passed as an array of strings. So the destination app will receive the identical array of strings.

yareach: Do we have any specs what params pt-mongodb-summary is to accept and what, to refuse?

percona-csalguero (approved):

client/client.go
```diff
@@ -535,6 +540,35 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
 	}
 }
 
+// argListFromMgDbParams creates an array of strings from the pointer to the parameters for pt-mongodb-sumamry
```
BupycHuk, line 543:
```suggestion
// argListFromMongoDBParams creates an array of strings from the pointer to the parameters for pt-mongodb-sumamry
```

BupycHuk (approved):

PR #221, opened 2021-01-13
PMM-6993 Run EXPLAIN queries in transaction

jprukner:
It might be destructive to use EXPLAIN in MySQL because stored
functions and procedures are being called.

Now, we wrap EXPLAIN in transaction so we can rollback any
changes after EXPLAIN is done.

### DIFF
--- a/actions/mysql_explain_action_test.go
+++ b/actions/mysql_explain_action_test.go
@@ -210,5 +210,57 @@ id |select_type |table |partitions |type |possible_keys |key  |key_len |ref  |ro
 			require.NoError(t, err)
 			checkCity(t)
 		})
+
+		t.Run("Stored function", func(t *testing.T) {
+			t.Parallel()
+
+			check := func(t *testing.T) {
+				t.Helper()
+				var count int
+				err := db.QueryRow("SELECT COUNT(*) FROM test_explain_table").Scan(&count)
+				require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


actions/mysql_explain_action_test.go
```diff
@@ -210,5 +210,42 @@ id |select_type |table |partitions |type |possible_keys |key  |key_len |ref  |ro
...
+				conn, err := db.Conn(ctx)
+				require.NoError(t, err)
+				defer conn.Close() //nolint:errcheck
+
+				_, err = conn.ExecContext(ctx, "DROP FUNCTION IF EXISTS cleanup")
+				require.NoError(t, err)
+				_, err = conn.ExecContext(ctx, `CREATE FUNCTION cleanup() RETURNS char(50) CHARSET latin1
+				DETERMINISTIC
+				BEGIN
+				delete from world.city;
```
BupycHuk, line 230: Let's make a little improvement and create a separate table to manipulate.

BupycHuk (changes requested):
Let's make tests more independent from each

actions/mysql_explain_action_test.go
jprukner: Done

BupycHuk (approved):

JiriCtvrtka (approved):

askomorokhov (approved):

PR #223, opened 2021-01-26
Fix timings for benchmarks

AlekSi:


BupycHuk:
Why these values were changed?

AlekSi:
It timeouts :)

BupycHuk (approved):

PR #230, opened 2021-02-05
PMM-5194 Fix locks

AlekSi:
The new protobuf package added mutexes to proto messages that should not be copied by a value.


.golangci.yml
```diff
@@ -6,8 +6,9 @@ linters-settings:
     packages:
       # use "github.com/pkg/errors" instead
       - errors
-      # use "github.com/golang/protobuf/proto" instead
+      # we use "github.com/golang/protobuf/proto" for now
```
BupycHuk, line 9: for now?

BupycHuk (approved):
How does this fixes locks?

BupycHuk:
ah, got it.

PR #243, opened 2021-03-11
PMM-7590 Jobs implemetation

artemgavrilov:
https://jira.percona.com/browse/PMM-7590

- [x] https://github.com/percona/pmm/pull/702
- [ ] https://github.com/percona/pmm-managed/pull/700

FB: Build: https://github.com/Percona-Lab/pmm-submodules/pull/1624


client/channel/actions.go
```diff
@@ -274,9 +274,9 @@ func (c *Channel) subscribe(id uint32) chan agentpb.ServerResponsePayload {
 	return ch
 }
 
-func (c *Channel) publish(id uint32, resp agentpb.ServerResponsePayload) {
+func (c *ActionsChannel) publish(id uint32, resp agentpb.ServerResponsePayload) {
```
BupycHuk, line 277: Actually this is the channel for everything.

artemgavrilov: Yep, I just don't find an appropriate name. Do you think it's better to leave it just `Channel`?

jobs/echo_job.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (j *echoJob) Type() string {
+	return "echo"
+}
+
+func (j *echoJob) Timeout() time.Duration {
+	return j.timeout
+}
+
+func (j *echoJob) Run(ctx context.Context, sender Sender) {
```
artemgavrilov, line 61: So idea is to let jobs send random messages back to the server when it's needed by job logic.

client/channel/actions.go
BupycHuk: Maybe I don't understand, but why did we create separate channel for jobs?

artemgavrilov: Because the current channel is semi-synchronous, you can start the action and subscribe for the result, but with jobs it can be hundreds of messages.

https://github.com/percona/pmm-agent/blob/0fb1bd754a83108224c8901b8fe96fb267f8e9b9/client/channel/channel.go#L68

artemgavrilov: I mean current channel implementation is not `just channel` it has some message handling logic that doesn't fit the jobs. Maybe I'm wrong, but I don't find a simple way to reuse the existing channel.

BupycHuk: we still can use the same mechanism that we use to collect qan metrics?

BupycHuk: What exact problem we have with the current implementation. Maybe I don't see something

artemgavrilov: Refactored, now existing channel reused for jobs

jobs/echo_job.go
```diff
@@ -0,0 +1,103 @@
...
+				Timestamp: ptypes.TimestampNow(),
+				Result: &agentpb.JobResult_Echo_{
+					Echo: &agentpb.JobResult_Echo{
+						Message: j.message,
+					},
+				},
+			},
+		})
+	case <-ctx.Done():
+		sender.SendResponse(&channel.AgentResponse{
```
BupycHuk, line 90: looks like we can implement it in `JobsRunner`, what do you think?

jobs/runner.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"context"
+	"runtime/pprof"
+	"sync"
+
+	"github.com/sirupsen/logrus"
+)
+
+type Runner struct {
```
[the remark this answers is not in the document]
BupycHuk: ^

jobs/runner.go
```diff
@@ -0,0 +1,113 @@
...
+	sender Sender
+
+	jobs        chan Job
+	runningJobs sync.WaitGroup
+
+	rw         sync.RWMutex
+	jobsCancel map[string]context.CancelFunc
+}
+
+func NewRunner(sender Sender) *Runner {
```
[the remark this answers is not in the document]
BupycHuk: ^

jobs/runner.go
```diff
@@ -0,0 +1,113 @@
...
+	return &Runner{
+		l:          logrus.WithField("component", "jobs-executor"),
+		sender:     sender,
+		jobs:       make(chan Job, 32), // TODO add constant
+		jobsCancel: make(map[string]context.CancelFunc),
+	}
+}
+
+func (e *Runner) Run(ctx context.Context) {
```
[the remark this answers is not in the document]
BupycHuk: ^

BupycHuk (changes requested):

agentlocal/agent_local.go
```diff
@@ -213,7 +213,7 @@ func (s *Server) runGRPCServer(ctx context.Context, listener net.Listener) {
 	}()
 	gRPCServer.GracefulStop()
 	shutdownCancel()
-	<-stopped // wait for Stop() to return
+	<-stopped // wait for StopJob() to return
```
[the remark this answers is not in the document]
artemgavrilov: Done

jobs/echo_job.go
```diff
@@ -0,0 +1,103 @@
...
+func (j *echoJob) Type() string {
+	return "echo"
+}
+
+func (j *echoJob) Timeout() time.Duration {
+	return j.timeout
+}
+
+func (j *echoJob) Run(ctx context.Context, sender Sender) {
+	sender.SendResponse(&channel.AgentResponse{
```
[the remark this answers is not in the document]
artemgavrilov: It should be simple `Send`, so we don't want any request, just `fire and forget`. So technically `SendRequsest` does what we need, but it is not about responses.

jobs/echo_job.go
artemgavrilov: Agree, done

jobs/echo_job.go
```diff
@@ -0,0 +1,103 @@
...
+		Payload: &agentpb.JobProgress{
+			JobId:     j.id,
+			Timestamp: ptypes.TimestampNow(),
+			Result: &agentpb.JobProgress_Echo_{
+				Echo: &agentpb.JobProgress_Echo{
+					Status: fmt.Sprintf("Echo job %s started", j.id),
```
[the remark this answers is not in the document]
artemgavrilov: Yes, for sure. But now it's not so clear what we need. I think it's better to add the required fields when we will have some real job.

jobs/echo_job.go
```diff
@@ -0,0 +1,103 @@
...
+				},
+			},
+		},
+	})
+	delay := time.NewTimer(j.delay)
+	defer delay.Stop()
+
+	select {
+	case <-delay.C:
+		sender.SendResponse(&channel.AgentResponse{
```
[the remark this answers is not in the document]
artemgavrilov: It actually can, job messages handled in a different way on the managed side.

jobs/echo_job.go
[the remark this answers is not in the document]
BupycHuk: Agree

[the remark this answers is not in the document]
artemgavrilov: Sorry, I confused words.

`SendResponse` does what we need here, just send a message
`SendRequest` expects some response back
That's why I used SendResponse, probably we can think about renaming, for example:

`SendResponse` -> `Send`
`SendRequest` -> `SendAndWaitForResponse` 
or something like this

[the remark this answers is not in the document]
artemgavrilov: @BupycHuk ^

client/client.go
```diff
@@ -343,16 +352,32 @@ func (c *Client) processChannelRequests(ctx context.Context) {
...
+			var job jobs.Job
+			switch j := p.Job.(type) {
+			case *agentpb.StartJob_Echo_:
+				job = jobs.NewEchoJob(
+					p.JobId,
+					p.Timeout.AsDuration(),
+					j.Echo.Message,
+					j.Echo.Delay.AsDuration(),
+				)
+			}
```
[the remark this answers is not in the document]
artemgavrilov: Yep, done

client/client.go
```diff
@@ -366,10 +399,10 @@ func (c *Client) processChannelRequests(ctx context.Context) {
 	}
 
 	if err := c.channel.Wait(); err != nil {
-		c.l.Debugf("Channel closed: %s.", err)
+		c.l.Debugf("ActionsChannel closed: %s.", err)
```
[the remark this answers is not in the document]
artemgavrilov: done

client/client.go
```diff
@@ -366,10 +399,10 @@ func (c *Client) processChannelRequests(ctx context.Context) {
...
 		return
 	}
-	c.l.Debug("Channel closed.")
+	c.l.Debug("ActionsChannel closed.")
```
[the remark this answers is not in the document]
artemgavrilov: done

jobs/runner.go
```diff
@@ -0,0 +1,148 @@
...
+func (r *Runner) Start(job Job) {
+	r.jobs <- job
+}
+
+// Stop stops running Job.
+func (r *Runner) Stop(id string) {
+	r.rw.RLock()
+	defer r.rw.RUnlock()
+	if cancel, ok := r.jobsCancel[id]; ok {
+		cancel()
```
Dasio, line 125: can we comment why we don't delete entry from map after u stop the job?

artemgavrilov: job removes itself from this map, see:
https://github.com/percona/pmm-agent/pull/243/files#diff-de1254a95ad6e1392d2a3d6a789b164c19e8ef47582685f2f6ec0bf001e49de9R144

Dasio: Yeah, I found it. But it isn't clear (at least for me).
My first thought was that after cancel is called, delete it from map.

artemgavrilov: Ok, will add a comment 👌

client/client.go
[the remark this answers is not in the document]
BupycHuk: I meant to not fail like we do for other cases, but return error response to pmm-managed. So we will not have to handle it on backend side each time.

jobs/echo_job.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Timeouts returns job timeout.
+func (j *echoJob) Timeout() time.Duration {
+	return j.timeout
+}
+
+// Run runs job.
+func (j *echoJob) Run(ctx context.Context, send Send) error {
+	j.l.Info("Job started.")
+	send(&agentpb.JobProgress{
```
BupycHuk, line 66: Do you handle that job progress messages can be sent in different order?

BupycHuk: let's use channels to handle it

jobs/runner.go
```diff
@@ -0,0 +1,152 @@
...
+// Sender provides method for sending message to PMM server.
+type Sender interface {
+	SendResponse(msg *channel.AgentResponse)
+}
+
+// Runner allows to execute jobs.
+type Runner struct {
+	l *logrus.Entry
+
+	sender Sender
```
BupycHuk, line 41: let's use channels instead of sender interface, it'll bring us more issues with race conditions.

BupycHuk: calling sendResponse makes data flow uncontrollable

BupycHuk: And harder to read and understand

jobs/runner.go
```diff
@@ -0,0 +1,152 @@
...
+			go pprof.Do(nCtx, pprof.Labels("jobID", jobID, "type", jobType), run)
+		case <-ctx.Done():
+			r.runningJobs.Wait() // wait for all jobs termination
+			return
+		}
+	}
+}
+
+func (r *Runner) send(payload agentpb.AgentResponsePayload) {
+	r.sender.SendResponse(&channel.AgentResponse{
```
BupycHuk, line 111: I would say it should be `Send` instead of `SendResponse`

BupycHuk (changes requested):

client/client.go
```diff
@@ -172,6 +176,10 @@ func (c *Client) Run(ctx context.Context) error {
 	// https://jira.percona.com/browse/PMM-4245
 
 	oneDone := make(chan struct{}, 3)
```
Dasio, line 178: should'nt this be bumped to 4?
Also why we are not using wait groups?

client/client.go
```diff
@@ -151,6 +153,8 @@ func (c *Client) Run(ctx context.Context) error {
 	c.channel = dialResult.channel
 	c.rw.Unlock()
 
+	c.jobsRunner = jobs.NewRunner(c.channel)
+
 	// Once the client is connected, ctx cancellation is ignored by it.
 	//
 	// We start three goroutines, and terminate the gRPC connection and exit Run when any of them exits:
```
Dasio, line 160: update comment

client/client.go
```diff
@@ -343,16 +353,43 @@ func (c *Client) processChannelRequests(ctx context.Context) {
...
 		case *agentpb.StopActionRequest:
-			c.runner.Stop(p.ActionId)
+			c.actionsRunner.Stop(p.ActionId)
 			responsePayload = new(agentpb.StopActionResponse)
 
 		case *agentpb.CheckConnectionRequest:
 			responsePayload = c.connectionChecker.Check(ctx, p, req.ID)
 
+		case *agentpb.StartJobRequest:
+			var job jobs.Job
```
Dasio, line 367: maybe also extract these code to function?
I don't like case inside case. But it's just my personal opinion.

Dasio (approved):

client/client.go
artemgavrilov: Yep, it should 👍 . Fixed.

Here is a trick, once one goroutine terminated this function returns immediately, and on upper-level context killed, and because of that other goroutines start terminating.

client/client.go
artemgavrilov: Yep, done

client/client.go
artemgavrilov: Done

oter (approved):

client/client.go
```diff
@@ -171,11 +174,19 @@ func (c *Client) Run(ctx context.Context) error {
 	// TODO Make 2 and 3 behave more like 1 - that seems to be simpler.
```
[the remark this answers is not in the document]
artemgavrilov: Mmm, I don't get it. What you want me to update? This description about what should be refactored in existing functions. In functions added it this PR nothing should be refactored (I hope so).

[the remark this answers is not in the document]
oter: sorry, I've added the comment on wrong line, I'm about lines 161-172 where described which goroutines are started, let's add processJobsResults and jobsRunner.Run()

jobs/runner.go
artemgavrilov: Done

jobs/runner.go
artemgavrilov: Renamed

jobs/echo_job.go
artemgavrilov: Done

client/client.go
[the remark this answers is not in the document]
BupycHuk: @artemgavrilov this is not resolved yet.

BupycHuk (approved):

client/client.go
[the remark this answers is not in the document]
artemgavrilov: Sorry, was triggered by the `default` word and missed the part about return. Done.

PR #244, opened 2021-03-11
PMM-7662: move to go modules

oter:
https://jira.percona.com/browse/PMM-7662


BupycHuk:
Tests are failed

Makefile
```diff
@@ -1,3 +1,16 @@
+BASE_PATH = $(shell pwd)
+BIN_PATH := $(BASE_PATH)/bin
+
+export PATH := $(BIN_PATH):$(PATH)
+
+GOLANGCI_LINT := $(BIN_PATH)/golangci-lint
```
Dasio, line 6: Is this necssary for each binary? We should agree on these, so we would be consistent across repositories

Dasio (approved):

Makefile
oter: it's discussable, but I see 2 pros in such way:
-  you don't need to specify `bin` folder
- a full path to binary is used

artemgavrilov: I personally prefer to not use such small variables for every binary that we use. It leads to variable hell and increases Makefile complexity. IMHO it's not a big deal to copy-paste `bin/something` in several places, but have really simple, easy-to-read makefile.

.github/workflows/go.yml
```diff
@@ -28,7 +28,6 @@ jobs:
     env:
       GOPATH: ${{ github.workspace }}
       WORKING_DIR: ${{ github.workspace }}/src/github.com/percona/pmm-agent
-      GO111MODULE: off #TODO: remove once we move to go modules https://jira.percona.com/browse/PMM-7106
```
artemgavrilov, line 31: Let's add go modules cache to CI
https://github.com/percona-platform/platform/blob/77210c15d1db136a396d9f382b97c170877410d0/.github/workflows/ci.yml#L66-L72

and cache cleaner 
https://github.com/percona-platform/platform/blob/main/.github/workflows/clean.yml

artemgavrilov (changes requested):

Makefile
BupycHuk: Agree with @artemgavrilov

BupycHuk:
@oter Tests are failing

Makefile
oter: as per discussion the variables removed

Makefile
```diff
@@ -1,3 +1,8 @@
+BASE_PATH = $(shell pwd)
+BIN_PATH := $(BASE_PATH)/bin
+
+export PATH := $(BIN_PATH):$(PATH)
+
```
artemgavrilov, line 5: Why do we need this?

.github/workflows/go.yml
oter: added

oter:
@BupycHuk tests are green now

Makefile
oter: are you about all the variables?

artemgavrilov: Yep, we have several new variables, but nothing changed in fact.

artemgavrilov: Discussed and resolved

.github/workflows/clean.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Enable Go build cache
+        uses: percona-platform/cache@v2
+        with:
+          path: ~/.cache/go-build
+          key: ${{ matrix.os }}-go-${{ matrix.go-version }}-build-${{ github.ref }}-${{ hashFiles('**') }}
+          restore-keys: |
+            ${{ matrix.os }}-go-${{ matrix.go-version }}-build-${{ github.ref }}-
+            ${{ matrix.os }}-go-${{ matrix.go-version }}-build-
+      - name: Clean Go modules cache
+        run: go clean -modcache
```
artemgavrilov, line 72: new line

artemgavrilov (approved):

.github/workflows/clean.yml
oter: added

BupycHuk (approved):

PR #256, opened 2021-04-30
PMM-7567 MongoDB backup

artemgavrilov:
https://jira.percona.com/browse/PMM-7567

- [x] https://github.com/percona/pmm/pull/726
- [x] https://github.com/percona/pmm-managed/pull/746


go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/percona/exporter_shared v0.7.2
 	github.com/percona/go-mysql v0.0.0-20200630114833-b77f37c0bfa2
 	github.com/percona/percona-toolkit v3.2.1+incompatible
-	github.com/percona/pmm v2.16.1-0.20210422091105-40d570e7d167+incompatible
+	github.com/percona/pmm v2.16.1-0.20210422114500-d75ab14f24c6+incompatible
```
artemgavrilov, line 29: TODO

client/client.go
```diff
@@ -446,6 +446,31 @@ func (c *Client) handleStartJobRequest(p *agentpb.StartJobRequest) error {
...
+				AccessKey:    cfg.S3Config.AccessKey,
+				SecretKey:    cfg.S3Config.SecretKey,
+				BucketName:   cfg.S3Config.BucketName,
+				BucketRegion: cfg.S3Config.BucketRegion,
+			}
+		default:
+			return errors.Errorf("unknown location config: %T", j.MongodbBackup.LocationConfig)
+		}
+
+		cfg := jobs.DatabaseConfig{
```
BupycHuk, line 465: ConnectionConfig? Credentials?
DatabaseConfig feels like something which contains database configuration.

jobs/mongodb_backup_job.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (j *MongoDBBackupJob) writePBMConfigFile() (string, error) {
+	tmp, err := ioutil.TempFile("", "pbm-config-*.yml")
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create pbm configuration file")
+	}
+
+	var conf struct {
+		Storage struct {
+			Typ string `yaml:"type"`
```
BupycHuk, line 227: is it intentional?

jobs/mongodb_backup_job.go
```diff
@@ -0,0 +1,261 @@
...
+
+	"github.com/golang/protobuf/ptypes"
+	"github.com/percona/pmm/api/agentpb"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	pbmBin = "pbm"
```
BupycHuk, line 37: it would be good to make path configurable via `pmm-agent.yaml` file

BupycHuk (changes requested):
What about TLS? do we support it? In this case I would suggest to pass ready URL from pmm-managed to control it there.

jobs/mongodb_backup_job.go
```diff
@@ -0,0 +1,261 @@
...
+
+	if err != nil {
+		return errors.Wrapf(err, "pbm config error: %s", string(output))
+	}
+
+	return nil
+}
+
+func (j *MongoDBBackupJob) writePBMConfigFile() (string, error) {
+	tmp, err := ioutil.TempFile("", "pbm-config-*.yml")
```
Dasio, line 220: can we defer Close here, so you don't need to close it each time when error occured?

pmm-agent-dev.yaml.bkp
```diff
@@ -0,0 +1,29 @@
+# Updated by `pmm-agent setup`.
```
Dasio, line 1: remove this file

Dasio (approved):

jobs/mongodb_backup_job.go
artemgavrilov: https://www.joeshaw.org/dont-defer-close-on-writable-files/

We need an error from `tmp.Close` when it's the last step, because it can mean that file is corrupted, but we need just to try to close a file in case of other errors.

pmm-agent-dev.yaml.bkp
artemgavrilov: 😱

jobs/mongodb_backup_job.go
artemgavrilov: Well, for now we expect that `pbm` is available in `PATH`. Later we will ship this binary with `pmm-agent`. If we want this configuration option now better to create separate ticket, because we already have hardcoded binaries for MySQL backup/restore.

jobs/mongodb_backup_job.go
artemgavrilov: It was while this field was private. Fixed.

artemgavrilov:
> What about TLS? do we support it? In this case I would suggest to pass ready URL from pmm-managed to control it there.

We expect that `pmm-agent` and `pbm-agent` started locally, on the same machine ad DB. So all connections except performed by backup tooling are local.

client/client.go
artemgavrilov: Renamed to `DBConnConfig`

BupycHuk (approved):

BupycHuk:
@artemgavrilov pmm-agent and mongodb server might be located on the same machine, but mongodb server might be configured to require TLS. So let's create a task for that.

PR #258, opened 2021-05-03
PMM-7917 pmm-agent high CPU usage for postgres

percona-csalguero:
When starting the agent or PostgreSQL connection comes back after a
disconnection, populating table names for the cache use too much CPU.

### DIFF
--- a/agents/postgres/pgstatstatements/pgstatstatements_test.go
+++ b/agents/postgres/pgstatstatements/pgstatstatements_test.go
@@ -279,7 +279,7 @@ func TestPGStatStatementsQAN(t *testing.T) {
 			Common: &agentpb.MetricsBucket_Common{
 				Fingerprint:         selectAllCitiesLong,
 				Database:            "pmm-agent",
-				Tables:              []string{"city"},
+				Tables:              []string{},
 				Username:            "pmm-agent",
 				AgentId:             "agent_id",
 				PeriodStartUnixSecs: 1554116340,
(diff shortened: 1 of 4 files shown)


JiriCtvrtka:
@cfsalguero please take a look, there is problem with query syntax.

JiriCtvrtka (approved):
LGTM, please check why one check failed on timeout.

BupycHuk (approved):
Can we add a few tests to makeBucket?

askomorokhov (approved):
Looks good to me! 👍

percona-csalguero:
@BupycHuk we are already testing `makeBuckets`.
[pgstatstatements_test.go#L177](https://github.com/percona/pmm-agent/blob/master/agents/postgres/pgstatstatements/pgstatstatements_test.go#L177) calls `getNewBuckets` which [calls makeBuckets](https://github.com/percona/pmm-agent/blob/master/agents/postgres/pgstatstatements/pgstatstatements.go#L187)

agents/postgres/pgstatstatements/pgstatstatements_test.go
```diff
@@ -322,7 +322,7 @@ func TestPGStatStatementsQAN(t *testing.T) {
 			Common: &agentpb.MetricsBucket_Common{
 				Fingerprint:         selectAllCitiesLong,
 				Database:            "pmm-agent",
-				Tables:              []string{"city"},
+				Tables:              []string{},
```
BupycHuk, line 325: then it's wrong test `agentpb.MetricsBucket_Common` should contain table name

percona-csalguero: It is part of the [AllCitiesTruncated](https://github.com/percona/pmm-agent/pull/258/files#diff-aa91ff13c274f4a1ba9eff1e32dd8bd3f4a8566b6bf4d1135408112bcf34d892R255) and the extract tables function fails since the query is not valid.
The logger shows `Can't extract table names from query SELECT /* AllCitiesTruncated:pgstatstatements */ * FROM city WHERE`
The [AllCities](https://github.com/percona/pmm-agent/pull/258/files#diff-aa91ff13c274f4a1ba9eff1e32dd8bd3f4a8566b6bf4d1135408112bcf34d892R191) subtest has the list of tables.

PR #260, opened 2021-05-06
PMM-6278 Explain DML queries

percona-csalguero:
Jira: [PMM-6278 Explain not showing for DELETE queries](https://jira.percona.com/browse/PMM-6278)

Only newer MySQL versions are able to run explain on DML queries but even if it is possible to run explain, it requieres more permissions than the ones we require for the connection user.
For example, to explain a `DROP TABLE` you should have drop permissions. 
This PR implements the same query transform funtions we have in Percona Toolkit and in PMM v1 that convert DML queries to the equivalent SELECTs when possible.
We only convert `INSERT`, `UPDATE` and `DELETE` queries. 
All other queries will be trated as usual thus, explain will return an error.


askomorokhov (approved):
Looks good to me 👍 
Do we need any update on UI side?

actions/query_transform.go
```diff
@@ -0,0 +1,125 @@
...
+*/
+func dmlToSelect(query string) string {
+	m := updateRe.FindStringSubmatch(query)
+	// > 2 because we need at least a table name and a list of fields
+	if len(m) > 2 {
+		return updateToSelect(m)
+	}
+
+	m = deleteRe.FindStringSubmatch(query)
+	if len(m) > 1 {
```
JiriCtvrtka, line 62: @percona-csalguero Maybe it is better to move these checks into methods itselfs. It will be validated also if you will call this method directly. What do you think? In all cases. It can save a lot of ifs.

actions/mysql_explain_action.go
```diff
@@ -34,8 +34,21 @@ import (
...
 }
 
+type explainResponse struct {
+	ExplainResult []byte `json:"explain_result"`
+	Query         string `json:"explained_query"`
+	IsDMLQuery    bool   `json:"is_dml"`
+}
+
+// ErrCannotEncodeExplainResponse cannot JSON encode the explain response.
+var ErrCannotEncodeExplainResponse = errors.New("cannot JSON encode the explain response")
```
JiriCtvrtka, line 50: Small one. It can be private for now.

JiriCtvrtka (commented):
Everything else looks good for me.

actions/query_transform.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return false
+}
+
+/*
+  MySQL version prior 5.6.3 cannot run explain on DML commands.
+  From the doc: http://dev.mysql.com/doc/refman/5.6/en/explain.html
+  "As of MySQL 5.6.3, permitted explainable statements for EXPLAIN are
+  SELECT, DELETE, INSERT, REPLACE, and UPDATE.
+  Before MySQL 5.6.3, SELECT is the only explainable statement."
```
BupycHuk, line 49: comment should be updated to match the reason why we convert DML queries

BupycHuk (changes requested):
I don't see related PR in pmm-managed or FE.

actions/query_transform.go
percona-csalguero: It is not validating, it is checking if each regular expression matches to choose which function to call.
if `updteRe` matches and returns at least 2 values then call `updateToSelect` and so on.

BupycHuk (approved):

PR #262, opened 2021-05-07
PMM-6552 Before register, check if config file is writable

Dasio:
https://jira.percona.com/browse/PMM-6552

Also when node already exists, inform user about `--force` option

FB: https://github.com/Percona-Lab/pmm-submodules/pull/1748


config/config_windows.go
```diff
@@ -0,0 +1,42 @@
...
+	"os"
+)
+
+// IsWritable tries to open and write comment to file.
+func IsWritable(path string) error {
+	file, err := os.OpenFile(path, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0640)
+	if err != nil {
+		return err
+	}
+	_, err = file.Write([]byte("# Write check\n"))
```
Dasio, line 33: maybe this write is not necessary?

config/config_windows.go
```diff
@@ -0,0 +1,42 @@
+// pmm-agent
```
BupycHuk, line 1: Do we support running pmm client in windows OS. As far as I know we don't.

commands/setup.go
```diff
@@ -62,6 +63,10 @@ func Setup() {
 	}
 
 	if !cfg.Setup.SkipRegistration {
+		if err := config.IsWritable(configFilepath); err != nil {
```
BupycHuk, line 66: we can extract it from this `if`. command may fail even with `--skip-registation` flag and it's better to have the same message in both cases.

BupycHuk (changes requested):

config/config_windows.go
Dasio: So let's delete both files (_unix,_windows), move to config.go and use just unix package?

BupycHuk: yeah

BupycHuk (approved):

commands/setup.go
```diff
@@ -61,6 +62,11 @@ func Setup() {
 		os.Exit(1)
 	}
 
+	if err := config.IsWritable(configFilepath); err != nil {
+		fmt.Printf("Config file %s is not writable: %s.\n", configFilepath, err)
```
artemgavrilov, line 66:
```suggestion
		fmt.Printf("Config file %s is not writable: %v.\n", configFilepath, err)
```

artemgavrilov (approved):

commands/setup.go
artemgavrilov: Technically both correct, but golang team uses `%v` in their examples:
https://blog.golang.org/error-handling-and-go

Also in our codebase `%v` used for errors printing.

PR #266, opened 2021-05-12
PMM-5917 check RELOAD privilege before log rotation

palash25:
https://jira.percona.com/browse/PMM-5917


agents/mysql/slowlog/slowlog.go
```diff
@@ -163,7 +164,25 @@ func (s *SlowLog) Run(ctx context.Context) {
 
 // recheck returns new slowlog information, and rotates slowlog file if needed.
 func (s *SlowLog) recheck(ctx context.Context) (newInfo *slowLogInfo) {
-	var err error
+	db, err := sql.Open("mysql", s.params.DSN)
+	if err != nil {
+		s.l.Errorf("Cannot open database connection: %s", err)
+		return
+	}
+	defer db.Close() //nolint:errcheck
```
oter, line 172: can we log the error?

agents/mysql/slowlog/slowlog.go
```diff
@@ -163,7 +164,25 @@ func (s *SlowLog) Run(ctx context.Context) {
 
 // recheck returns new slowlog information, and rotates slowlog file if needed.
 func (s *SlowLog) recheck(ctx context.Context) (newInfo *slowLogInfo) {
-	var err error
+	db, err := sql.Open("mysql", s.params.DSN)
+	if err != nil {
+		s.l.Errorf("Cannot open database connection: %s", err)
```
oter, line 169: maybe we can return such errors and log them on the caller's side?

percona-csalguero (approved):

agents/mysql/slowlog/slowlog.go
palash25: disagree,  if we return the error and still just log it then its not much different from the current code. we will also have to change the function signature and the return statements everywhere in `recheck` better to prefer a smaller code change over a slightly bigger one if both do the same thing

jprukner (approved):
Were docs updated to include this privileges?

oter (approved):

palash25:
> Were docs updated to include the privilege?

the docs were already there, it was this check that was missing

PR #269, opened 2021-05-19
PMM-8001 pg_stat_monitor version error handling.

JiriCtvrtka:


askomorokhov (approved):

percona-csalguero (approved):

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -116,14 +116,37 @@ func newPgStatMonitorQAN(q *reform.Querier, dbCloser io.Closer, agentID string,
...
+
+func parsePGMonitorVersion(version string) (float64, error) {
+	split := strings.Split(version, ".")
+	switch len(split) {
+	case 1:
+		return strconv.ParseFloat(split[0], 64)
+	case 2:
+		if len(split[1]) == 1 {
+			split[1] = "0" + split[1]
+		}
```
BupycHuk, line 137: what if version will go up to `0.20.0`?

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -116,14 +116,37 @@ func newPgStatMonitorQAN(q *reform.Querier, dbCloser io.Closer, agentID string,
...
+
+	return parsePGMonitorVersion(v)
+}
+
+func parsePGMonitorVersion(version string) (float64, error) {
+	split := strings.Split(version, ".")
+	switch len(split) {
+	case 1:
+		return strconv.ParseFloat(split[0], 64)
+	case 2:
```
BupycHuk, line 134: did you ask pg_stat_monitor team if it's possible to have less then 3 numbers in version?

BupycHuk (changes requested):
it's better to use https://github.com/hashicorp/go-version or the one from github.com/percona/pmm to compare versions instead of float numbers.

agents/postgres/pgstatmonitor/pgstatmonitor.go
JiriCtvrtka: For 0.20.0 it will be 0.2 and for 0.2.0 it will be 0.02. Please check similar cases in test here:
https://github.com/percona/pmm-agent/blob/PMM-8001-pg-stat-monitor-error-handle/agents/postgres/pgstatmonitor/pgstatmonitor_test.go#L481:L487

agents/postgres/pgstatmonitor/pgstatmonitor.go
JiriCtvrtka: No, I will ping them for answer.

EngineeredVirus (commented):
pg_stat_monitor version scheme has 3 independent parts that are being used to track major, minor, and patch versions: X.Y.Z

These versions should be separately compared, e.g. when comparing X1.Y1.Z1 and X2.Y2.Z2, the logic should follow:

If X1 is greater than X2, then X1.Y1.Z1 is the newer version.
If X1 is less than X2, then X2.Y2.Z2 is the newer version.

If X1 is equal to X2, then
- If Y1 is great than Y2, then X1.Y1.Z1 is the newer version.
- If Y1 is less than Y2, then X2.Y2.Z2 is the newer version.
- If Y1 is equal to Y2, then 
--- If Z1 is great than Z2, then X1.Y1.Z1 is the newer version.
--- If Z1 is less than Z2, then X2.Y2.Z2 is the newer version.
--- If Z1 is equal to Z2, then both versions are the same.

For comparison, you may use integers rather than floats.

agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"gopkg.in/reform.v1"
```
BupycHuk, line 29: please remove blank line

agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -83,19 +85,36 @@ func (ssc *statMonitorCache) getStatMonitorExtended(ctx context.Context, q *refo
...
 	switch {
-	case pgMonitorVersion >= 0.9:
+	case v.GreaterThanOrEqual(v09):
 		view = pgStatMonitor09View
 		row = &pgStatMonitor09{}
-	case pgMonitorVersion >= 0.8:
+	case v.GreaterThanOrEqual(v08):
 		view = pgStatMonitor08View
 		row = &pgStatMonitor08{}
 	default:
```
BupycHuk, line 120: let's extract all this to a separate method, so we can wrap error or use `view` and `row` returned from it.

agents/postgres/pgstatmonitor/stat_monitor_cache.go
JiriCtvrtka: Yes, removed.

agents/postgres/pgstatmonitor/stat_monitor_cache.go
JiriCtvrtka: Extracted to func.

BupycHuk (approved):

PR #276, opened 2021-06-17
PMM-8160: Versions compatibility

oter:
https://jira.percona.com/browse/PMM-8160

API PR: https://github.com/percona/pmm/pull/755
managed PR: https://github.com/percona/pmm-managed/pull/785

### DIFF
--- a/client/version.go
+++ b/client/version.go
@@ -24,7 +24,7 @@ const (
 )
 
 var (
-	mysqlVersionRegexp = regexp.MustCompile("^.*Ver ([!-~]*).*$")
+	mysqlVersionRegexp      = regexp.MustCompile("^.*Ver ([!-~]*).*$")
 	xtrabackupVersionRegexp = regexp.MustCompile("^xtrabackup version ([!-~]*).*$")
 )
 
--- a/client/version.go
+++ b/client/version.go
@@ -36,8 +36,8 @@ const (
 )
 
 var (
-	mysqlVersionRegexp      = regexp.MustCompile("^.*Ver ([!-~]*).*$")
-	xtrabackupVersionRegexp = regexp.MustCompile("^xtrabackup version ([!-~]*).*$")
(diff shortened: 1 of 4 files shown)


Dasio (approved):

client/version.go
```diff
@@ -0,0 +1,102 @@
+// pmm-agent
```
BupycHuk, line 1: this file should be separate structure.

client/version.go
```diff
@@ -0,0 +1,102 @@
...
+	mysqlBin            = "mysql"
+	xtrabackupBin       = "xtrabackup"
+)
+
+var (
+	mysqlVersionRegexp      = regexp.MustCompile("^.*Ver ([!-~]*).*")
+	xtrabackupVersionRegexp = regexp.MustCompile("^xtrabackup version ([!-~]*).*")
+)
+
+func (c *Client) localMySQLVersion() (string, error) {
```
BupycHuk, line 43: `mysqlClientBinaryVersion`?

BupycHuk (changes requested):

client/version.go
oter: could you please add more details.

oter: could you please add more details.

client/version.go
oter: I could rename it just to `mysqlVersion`. What do you think?

client/version.go
BupycHuk: I mean that `localMySQLVersion`, `xtrabackupVersion` methods doesn't look like it should be part of `Client` structure, my suggestion to make it separate function like we did for `CheckConnection`

oter: I see, will do.

client/version.go
BupycHuk: What do we check here? Version of MySQL client installed locally, aren't we?

oter: I need to check mysql server version instead of client  (`mysqld` instead of `mysql`) 🤦 Thanks for pointing this out!

PR #284, opened 2021-07-30
PMM-8300 Add load for pg_stat_monitor.

askomorokhov:
https://jira.percona.com/browse/PMM-8300

to create load run `docker-compose -f docker-compose-pg-load.yml up postgres-load`


percona-csalguero (changes requested):

utils/tests/db.go
```diff
@@ -17,21 +17,24 @@ package tests
...
 )
 
+var testMutex sync.Mutex
+
 // waitForFixtures waits up to 30 seconds to database fixtures (test_db) to be loaded.
 func waitForFixtures(tb testing.TB, db *sql.DB) {
 	var id int
 	var err error
 	for i := 0; i < 30; i++ {
+		testMutex.Lock()
```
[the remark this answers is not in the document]
JiriCtvrtka: @percona-csalguero I am trying to fix build, because since Andrii created this PR build were always failed. We have data racing in tests and it happening on more places.I will mark it as draft once more until build will be fixed.

docker-compose-pg-load.yml
```diff
@@ -0,0 +1,50 @@
+---
+ # docker-compose -f docker-compose-pg-load.yml up postgres-load
+version: '3.7'
+
+services:
+  postgres-pgmonitor:
```
BupycHuk, line 6: maybe it's better to include it to default docker-compose and add some instructions how to use it.

BupycHuk: I can't understand how we are going to use it. What's use case?

JiriCtvrtka: Usage is docker-compose -f docker-compose-pg-load.yml up postgres-load. Then let it running and it generating required fields in pg_stat_monitor. Add to monitoring and see new metrics filled here. It is same how Andrii did it, I just separate service pg with monitor to this file to avoid problems with test.

BupycHuk (approved):

percona-csalguero (approved):

PR #287, opened 2021-08-18
PMM-8463 Fix process hanging if 2nd process unexpected exits

Dasio:
https://jira.percona.com/browse/PMM-8463

If 2nd process exits without reading stdout from the 1st process. Then it will hangs forever.


jobs/mysql_backup_job.go
```diff
@@ -154,25 +163,14 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
...
-	if err := xtrabackupCmd.Start(); err != nil {
-		return wrapError(err)
+		if err := xbcloudCmd.Wait(); err != nil {
+			_ = xtrabackupStdout.Close()
+			return wrapError(err)
+		}
 	}
 
-	if err := xtrabackupCmd.Wait(); err != nil {
+	err = <-xtrabackupErr
```
BupycHuk, line 172: it won't come to this line if xbcloudcmd will fail, so we will have one goroutine which will hang.

jobs/mysql_backup_job.go
```diff
@@ -154,25 +163,14 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
...
-			} else {
-				rerr = wrapError(err)
-			}
-		}()
-	}
-
-	if err := xtrabackupCmd.Start(); err != nil {
-		return wrapError(err)
+		if err := xbcloudCmd.Wait(); err != nil {
+			_ = xtrabackupStdout.Close()
```
BupycHuk, line 167: so, closing stdout is a key, maybe that change be enough?
I think better option would be to terminate xtrabackup instead of closing stdout.

BupycHuk (changes requested):

jobs/mysql_restore_job.go
```diff
@@ -167,30 +167,29 @@ func restoreMySQLFromS3(
...
-			return
-		}
 
-		if rerr != nil {
-			rerr = errors.Wrapf(rerr, "xbcloud wait error: %s", err)
-		} else {
-			rerr = errors.Wrap(wrapError(err), "xbcloud wait failed")
-		}
+	go func() {
+		xbcloudErr <- xbcloudCmd.Wait()
```
oter, line 176: the documentation says it is incorrect to call Wait before all reads from the pipe have been completed.
https://pkg.go.dev/os/exec#Cmd.StdoutPipe

Probably a better option is to use io.Pipe

jobs/mysql_backup_job.go
Dasio: When I was testing it before changes. It would hang on both `Wait()`
With this changes without stdout close, only one hangs. With close, no one hangs.

BupycHuk: Don't we make xtrabackup fail by closing stdout? maybe it's better to terminate the process, since it's what we want to do

jobs/mysql_backup_job.go
Dasio: I tried it again with `io.Pipe` and goroutine exits properly.

oter (approved):

oter:
@BupycHuk or @nikita-b  CI/check step is failed, can you fix it one more time?

nikita-b:
Steps with `tip` version Go should be optional... @BupycHuk Am I right?

BupycHuk:
@oter you can't merge because I requested changes, that's the reason

jobs/mysql_backup_job.go
```diff
@@ -146,43 +173,42 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
 			errBackupBuffer.String(), outBuffer.String(), errCloudBuffer.String())
 	}
 
+	if err := xtrabackupCmd.Start(); err != nil {
```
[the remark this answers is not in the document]
oter: should be no, because the pipe will be write-blocked, until something starts reading from the pipe.

jobs/mysql_restore_job.go
```diff
@@ -168,28 +223,34 @@ func restoreMySQLFromS3(
...
 
 	if err := xbstreamCmd.Start(); err != nil {
+		cancel()
 		return errors.Wrap(wrapError(err), "xbstream start failed")
 	}
-
-	if err := xbstreamCmd.Wait(); err != nil {
-		return errors.Wrap(wrapError(err), "xbstream wait failed")
-	}
+	defer func() {
```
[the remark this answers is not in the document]
artemgavrilov: The intention was to have the same "blocks" where you have `start` + `defer wait`, so you can move such blocks of code without breaking the pipeline. I.e. you can copy-paste line 225-238, change change command and insert it somewhere in this job and it will work without any problems.

jobs/mysql_backup_job.go
```diff
@@ -96,7 +116,14 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
 		}
 	}
 
-	xtrabackupCmd := exec.CommandContext(ctx, xtrabackupBin, "--compress", "--backup") // #nosec G204
+	return nil
+}
+
+func (j *MySQLBackupJob) backup(pipeCtx context.Context) (rerr error) {
```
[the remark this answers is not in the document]
artemgavrilov: True

artemgavrilov (approved):

Dasio:
@BupycHuk are comment from Artem and Max enough? Or do you still propose some changes?

BupycHuk (approved):

PR #288, opened 2021-08-18
PMM-8463 Fix commands pipelining

artemgavrilov:
https://jira.percona.com/browse/PMM-8463

### DIFF
--- a/jobs/mysql_backup_job.go
+++ b/jobs/mysql_backup_job.go
@@ -81,10 +81,27 @@ func (j *MySQLBackupJob) Timeout() time.Duration {
 }
 
 // Run starts Job execution.
-func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+func (j *MySQLBackupJob) Run(ctx context.Context, send Send) error {
+	if err := j.binariesInstalled(); err != nil {
+		return errors.WithStack(err)
+	}
 
+	if err := j.backup(ctx); err != nil {
+		return errors.WithStack(err)
+	}
+
+	send(&agentpb.JobResult{
+		JobId:     j.id,
+		Timestamp: ptypes.TimestampNow(),
+		Result: &agentpb.JobResult_MysqlBackup{
+			MysqlBackup: &agentpb.JobResult_MySQLBackup{},
+		},
+	})
+
+	return nil
+}
+
+func (j *MySQLBackupJob) binariesInstalled() error {
 	if _, err := exec.LookPath(xtrabackupBin); err != nil {
 		return errors.Wrapf(err, "lookpath: %s", xtrabackupBin)
 	}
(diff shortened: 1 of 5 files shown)


jobs/mysql_restore_job.go
```diff
@@ -330,17 +332,6 @@ func restoreBackup(ctx context.Context, backupDirectory, mySQLDirectory string)
...
-	defer func(start time.Time) {
-		entry := j.l.WithField("duration", time.Since(start).String())
-
-		if rerr != nil {
-			entry.Error("MySQL restore finished with error")
-		} else {
-			entry.Info("MySQL restore finished")
-		}
-	}(time.Now())
-
```
artemgavrilov, line 343: We have it in runner

jobs/mysql_backup_job.go
```diff
@@ -96,7 +116,14 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
 		}
 	}
 
-	xtrabackupCmd := exec.CommandContext(ctx, xtrabackupBin, "--compress", "--backup") // #nosec G204
+	return nil
+}
+
+func (j *MySQLBackupJob) backup(pipeCtx context.Context) (rerr error) {
```
artemgavrilov, line 122: We should wrap pipeline-related commands to function to prevent sending job result message. I.e we need all defers with `cmd.Wait` to finish successfully before sending job result. 

This is important because one failed command will not break whole pipeline, for example `cat xx | echo message` cat will exit with error code, but echo will print `message` and will terminate successfully.

jobs/mysql_backup_job.go
```diff
@@ -96,7 +116,14 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
...
+	pipeCtx, cancel := context.WithCancel(pipeCtx)
+	defer cancel()
```
artemgavrilov, line 124: The main idea is to kill all commands in the pipeline when one of them is failed.

artemgavrilov: Also `cmd.Wait` order matters. Imagine situation where you have `cmd1 | cmd2`, so in your code you should with cmd2 completion and then cmd1.

jobs/mysql_backup_job.go
```diff
@@ -146,43 +173,42 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
...
+			}
+		}
+	}()
+
 	if xbcloudCmd != nil {
 		xbcloudCmd.Stdin = xtrabackupStdout
 		xbcloudCmd.Stdout = &outBuffer
 		xbcloudCmd.Stderr = &errCloudBuffer
 		if err := xbcloudCmd.Start(); err != nil {
+			cancel()
```
Dasio, line 197: the same here. Because firstly we start both process and just afterwards we call `Wait()` so this doesn't seems necessary?

jobs/mysql_backup_job.go
```diff
@@ -146,43 +173,42 @@ func (j *MySQLBackupJob) Run(ctx context.Context, send Send) (rerr error) {
 			errBackupBuffer.String(), outBuffer.String(), errCloudBuffer.String())
 	}
 
+	if err := xtrabackupCmd.Start(); err != nil {
+		cancel()
```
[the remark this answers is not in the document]
artemgavrilov: Here I added `cancel` to prvent errors during copy-paste. So technically it's not necessary here, but it you will add some commands before this command it will work anyway.

jobs/mysql_backup_job.go
artemgavrilov: Here it's required, because if we failed to start `xbcloud`  we invoke return, and that triggers defer with `xtrabackupCmd.Wait()`  where we will stuck.

artemgavrilov:
Changes are moved to the https://github.com/percona/pmm-agent/pull/287

PR #294, opened 2021-10-01
PMM-8843 handle truncation by pg_stat_monitor

jprukner:
The query can be truncated by pg_stat_monitor. If we fail to normalize truncated query, we check it's elevel to see if query was successful. If it was successful, we can say the query is truncated by the pg_stat_monitor. If it was not successful, there is a real syntax error in the query - we print the warning.

### DIFF
--- a/agents/postgres/pgstatmonitor/stat_monitor_cache.go
+++ b/agents/postgres/pgstatmonitor/stat_monitor_cache.go
@@ -150,16 +150,30 @@ func (ssc *statMonitorCache) getStatMonitorExtended(ctx context.Context, q *refo
 			example := ""
 			if !normalizedQuery {
 				example = c.Query
-				fingerprint = ssc.generateFingerprint(c.Query)
+				fingerprint, err = ssc.generateFingerprint(c.Query)
 			}
-			var isTruncated bool
-			c.Fingerprint, isTruncated = truncate.Query(fingerprint)
-			if isTruncated {
-				c.IsQueryTruncated = isTruncated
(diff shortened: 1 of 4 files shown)


JiriCtvrtka (approved):

percona-csalguero (approved):

agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -150,16 +150,30 @@ func (ssc *statMonitorCache) getStatMonitorExtended(ctx context.Context, q *refo
...
+				if c.pgStatMonitor.Elevel != 0 {
+					c.IsQueryTruncated = false
+					ssc.l.Warnf("generating fingerprint failed for query with id %v: %v", c.QueryID, err)
+				} else {
+					// We failed to normalize the query but the query was executed successfully.
+					// That means the query was truncated by pg_stat_monitor. Let's state the query was truncated.
+					c.IsQueryTruncated = true
+				}
+				c.Example = c.Query
+				c.Fingerprint = c.Query
```
BupycHuk, line 166: we should discuss it with @rnovikovP, since we show query as is which might be not secure.

rnovikovP: I'm ok with queries in logs when it's an error for them. 
We'll have the log level controlled from the PMM side, so for some secured env. this will be possible to OFF

BupycHuk (approved):

PR #301, opened 2021-10-20
PMM-8588 Histogram.

JiriCtvrtka:


agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -19,13 +19,17 @@ package pgstatmonitor
...
 	"fmt"
 	"io"
+	"strconv"
 	"time"
 
+	"github.com/lib/pq"
 	_ "github.com/lib/pq" // register SQL driver.
 	"github.com/percona/pmm/api/agentpb"
 	"github.com/percona/pmm/api/inventorypb"
+	"github.com/percona/pmm/api/qanpb"
```
BupycHuk, line 32: it's better to not import qanpb packages in pmm-agent, right now we can change internal implementation of PMM Server without having any issues with backward compatibility. As I mentioned in `pmm` PR, we can use just a `map` for histograms.

BupycHuk (changes requested):

agents/postgres/pgstatmonitor/pgstatmonitor.go
JiriCtvrtka: @BupycHuk Then marshalling will be done on QAN-API2 side, is this OK?

BupycHuk: if we will use map, then yes, but if we will use separate message type, then best option for us is to convert from pmm-agent type to qan-api2 type in pmm-managed.

BupycHuk:
And please fix CI

agents/postgres/pgstatmonitor/pgstatmonitor.go
JiriCtvrtka: Map will sort keys and this is problem here. I will do it by separate message type.

JiriCtvrtka:
About CI today here is problem with them. They often stuck on docker compose for long time (like here https://github.com/percona/pmm-agent/runs/4082638373?check_suite_focus=true), or error with timeout. @nikita-b should create ticket for IT about this. I tried several reruns.

BupycHuk:
I see this
```
2021/11/02 17:29:07 internal error: package "bytes" without types was imported from "github.com/percona/pmm-agent/agentlocal"
agentlocal/deps.go:26: running "mockery": exit status 1
2021/11/02 17:29:16 internal error: package "context" without types was imported from "github.com/percona/pmm-agent/client"
client/deps.go:24: running "mockery": exit status 1
2021/11/02 17:29:16 internal error: package "context" without types was imported from "github.com/percona/pmm-agent/versioner"
versioner/versioner.go:51: running "mockery": exit status 1
```

BupycHuk (approved):

JiriCtvrtka:
About failing test, localy it is fine. Also I tried create PR from 2.22 branch and it is same
https://github.com/percona/pmm-agent/pull/302

PR #310, opened 2022-01-05
PMM-8141 Recreate VM config on restart.

JiriCtvrtka:


percona-csalguero (approved):

agents/process/process.go
```diff
@@ -174,6 +179,14 @@ func (p *Process) toWaiting() {
 	defer t.Stop()
 	select {
 	case <-t.C:
+		// VM_AGENT need recreate config file in temp dir.
+		if p.params.Type == inventorypb.AgentType_VM_AGENT {
+			_, err := p.params.TemplateRenderer.RenderFiles(p.params.TemplateParams)
```
ShashankSinha252, line 184: Is it safe to regenerate config without checking if it is required?

JiriCtvrtka: We regenerate config on every start anyway. So do this during restart should not change anything. We can check on restart if file exists and if yes then dont regenerate new one (but this will not ensure that config is OK). I think it is more clear to have same behavior like in start process. What do you think?

ShashankSinha252: Considering the above, this approach sounds okay.

ShashankSinha252 (approved):

BupycHuk (approved):

PR #316, opened 2022-01-27
PMM-5279 Limit caches in size.

PavelKhripkov:
https://jira.percona.com/browse/PMM-5279

https://github.com/Percona-Lab/pmm-submodules/pull/2314


idoqo (commented):
Looks good to me, and interesting use of reflection too!

Also, I think you could check out some of the bot comments too - I think some of them are valid.

idoqo (approved):

agents/cache/cache.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		c.l.Debugf("Cache size exceeded the limit of %d items and the oldest values were trimmed. "+
+			"Now the oldest query in the cache is of time %s",
+			c.sizeLimit, itemList[overSize].t.UTC().Format("2006-01-02T15:04:05Z"))
+	}
+	return nil
+}
+
+// Stats returns Cache statistics.
+func (c *Cache) Stats() Stats {
```
artemgavrilov, line 137: As I see this function is used only for debug log, also this debug log is always printed after `Refresh` function. So maybe it makes sense just to move log inside `Refresh` function and make this method private.

agents/cache/cache.go
```diff
@@ -0,0 +1,194 @@
...
+		addedN:   c.addedN,
+		removedN: c.removedN,
+		oldest:   oldest,
+		newest:   newest,
+	}
+}
+
+// Len returns the current number of elements in Cache.
+func (c *Cache) Len() int {
+	return len(c.items)
```
artemgavrilov, line 164: `len` should be invoked under a mutex.

agents/cache/cache.go
```diff
@@ -0,0 +1,194 @@
...
+		updatedN: c.updatedN,
+		addedN:   c.addedN,
+		removedN: c.removedN,
+		oldest:   oldest,
+		newest:   newest,
+	}
+}
+
+// Len returns the current number of elements in Cache.
+func (c *Cache) Len() int {
```
artemgavrilov, line 163: Why it's a separate method, not just another field in `Stats`?

agents/cache/cache.go
```diff
@@ -0,0 +1,194 @@
...
+	typ       reflect.Type
+	items     statementsMap
+	added     statementsAddedMap
+	retain    time.Duration
+	sizeLimit uint
+	l         *logrus.Entry
+	rw        sync.RWMutex
+	updatedN  uint
+	addedN    uint
+	removedN  uint
```
artemgavrilov, line 48: Can we use Prometheus metrics for that? I think we should.

artemgavrilov (changes requested):
I have several thoughts:
1) Type safety is important stuff that protects us from mistakes during compile time. So one possible solution to have generic cache and type safety is to write simple wrappers for each type. 
```
type HistoryCache struct {
    cache *Cache
}
func (c *HistoryCache)Set(items []*historyItem) {
    c.cache.set(items)
}
```

2) We need a very basic cache with TTL, probably we can fit some solution on github instead of implementing it.

3) Generics will be released soon, probably it makes sense to wait for them.

agents/mongodb/mongodb.go
```diff
@@ -106,3 +107,18 @@ type Profiler interface {
...
+
+// Collect implement prometheus.Collector.
+func (m *MongoDB) Collect(ch chan<- prometheus.Metric) {
+	// not implemented
+}
+
+// check interfaces
+var (
+	_ prometheus.Collector = (*MongoDB)(nil)
+)
```
artemgavrilov, line 124: Why you added it if it's not implemented?

agents/mysql/perfschema/perfschema.go
```diff
@@ -131,21 +139,31 @@ func New(params *Params, l *logrus.Entry) (*PerfSchema, error) {
...
 	}
-	return newPerfSchema(newParams), nil
+	return newPerfSchema(newParams)
 }
 
-func newPerfSchema(params *newPerfSchemaParams) *PerfSchema {
+func newPerfSchema(params *newPerfSchemaParams) (*PerfSchema, error) {
+	historyCache, err := newHistoryCache(historyMap{}, retainHistory, historyCacheSize, params.LogEntry)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create cache: %w", err)
```
artemgavrilov, line 148: Lets use `errors.Wrap`, it adds stacktraces

agents/mysql/slowlog/slowlog.go
```diff
@@ -673,3 +674,18 @@ func errListsToMap(k, v []uint64) map[uint64]uint64 {
...
+
+// Collect implement prometheus.Collector.
+func (s *SlowLog) Collect(ch chan<- prometheus.Metric) {
+	// not implemented
+}
+
+// check interfaces
+var (
+	_ prometheus.Collector = (*SlowLog)(nil)
+)
```
artemgavrilov, line 691: Same, why?

agents/cache/utils.go
```diff
@@ -0,0 +1,75 @@
...
+
+// MetricsFromStats builds Prometheus metrics from cache.Stats.
+func MetricsFromStats(stats Stats, agentID string, cacheType string) []prometheus.Metric {
+	metrics := []prometheus.Metric{
+		prometheus.MustNewConstMetric(mCurrentDesc, prometheus.GaugeValue, float64(stats.Current), agentID, cacheType),
+		prometheus.MustNewConstMetric(mUpdatedNDesc, prometheus.CounterValue, float64(stats.UpdatedN), agentID, cacheType),
+		prometheus.MustNewConstMetric(mAddedNDesc, prometheus.CounterValue, float64(stats.AddedN), agentID, cacheType),
+		prometheus.MustNewConstMetric(mRemovedNDesc, prometheus.CounterValue, float64(stats.RemovedN), agentID, cacheType),
+		prometheus.MustNewConstMetric(mOldestDesc, prometheus.CounterValue, float64(stats.Oldest.Unix()), agentID, cacheType),
+		prometheus.MustNewConstMetric(mNewestDesc, prometheus.CounterValue, float64(stats.Newest.Unix()), agentID, cacheType),
```
artemgavrilov, line 28: Let's just move it inside `Collect` function, like we do it here: https://github.com/percona/pmm-managed/blob/1f20300018046bb68eb18878dab846a24e477ecc/services/agents/registry.go#L410

agents/cache/utils.go
```diff
@@ -0,0 +1,75 @@
...
+		[]string{"agent_id", "cache_type"},
+		nil,
+	)
+	mNewestDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(prometheusNamespace, prometheusSubsystem, "newest"),
+		"Timestamp of newest row in cache.",
+		[]string{"agent_id", "cache_type"},
+		nil,
+	)
+)
```
artemgavrilov, line 75: And this lets move to the `cache.go`

artemgavrilov (changes requested):

artemgavrilov (approved):

BupycHuk (commented):
Could you check how all these reflection affects performance?
Can we use interfaces instead of reflection?

agents/cache/cache.go
```diff
@@ -0,0 +1,179 @@
...
+	defer c.rw.Unlock()
+	now := time.Now()
+	var wasTrimmed bool
+
+	var next *list.Element
+	for e := c.itemsList.Front(); e != nil && now.Sub(e.Value.(*cacheItem).added) > c.retain; e = next {
+		c.removedN++
+		next = e.Next()
+		delete(c.items, c.itemsList.Remove(e).(*cacheItem).key)
+	}
```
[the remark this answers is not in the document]
PavelKhripkov: We use doubly linked list to keep items sorted by add time so we don't need to iterate over all items to delete the outdated ones.

agents/cache/cache.go
```diff
@@ -0,0 +1,179 @@
...
+	items     map[interface{}]*list.Element
+	itemsList *list.List
+	updatedN  uint
+	addedN    uint
+	removedN  uint
+}
+
+// cacheItem is an element stored in Cache
+type cacheItem struct {
+	key   interface{}
```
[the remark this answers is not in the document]
PavelKhripkov: > do we really need this field?
We need it for backward relation from List to map.

> what types can be used as key? I though it's always string in our cases.
It would be much easier case and we would use one of existing open source lib. But we also use int64 keys.

agents/cache/cache_test.go
```diff
@@ -0,0 +1,172 @@
...
+		assert.True(t, reflect.DeepEqual(actual, expected))
+		assert.Equal(t, uint(6), stats.Current)
+		assert.Equal(t, uint(4), stats.UpdatedN)
+		assert.Equal(t, uint(6), stats.AddedN)
+		assert.Equal(t, uint(0), stats.RemovedN)
+		assert.InDelta(t, 0, int(stats.Oldest.Sub(now1).Seconds()), 0.01)
+		assert.InDelta(t, 0, int(stats.Newest.Sub(now2).Seconds()), 0.01)
+	})
+
+	t.Run("ReachesTimeLimit", func(t *testing.T) {
```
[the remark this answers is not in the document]
PavelKhripkov: It's done via 'stats'

PR #332, opened 2022-03-17
Easify use CONTRIBUTING.md

YaroslavPodorvanov:
PMM-0

Build: SUBMODULES-0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,15 @@ release:                        ## Build static pmm-agent release binary (Linux 
 	$(PMM_RELEASE_PATH)/pmm-agent --version
 	-ldd $(PMM_RELEASE_PATH)/pmm-agent
 
+download-exporters:
+	cd .. && git clone https://github.com/percona/node_exporter || cd ./node_exporter && git pull
+	cd .. && git clone https://github.com/percona/mysqld_exporter || cd ./mysqld_exporter && git pull
+	cd .. && git clone https://github.com/percona/rds_exporter || cd ./rds_exporter && git pull
+	cd .. && git clone https://github.com/percona/postgres_exporter || cd ./postgres_exporter && git pull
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -28,6 +28,15 @@ release:                        ## Build static pmm-agent release binary (Linux
...
 	-ldd $(PMM_RELEASE_PATH)/pmm-agent
 
+download-exporters:
+	cd .. && git clone https://github.com/percona/node_exporter || cd ./node_exporter && git pull
+	cd .. && git clone https://github.com/percona/mysqld_exporter || cd ./mysqld_exporter && git pull
+	cd .. && git clone https://github.com/percona/rds_exporter || cd ./rds_exporter && git pull
+	cd .. && git clone https://github.com/percona/postgres_exporter || cd ./postgres_exporter && git pull
+	cd .. && git clone https://github.com/percona/mongodb_exporter || cd ./mongodb_exporter && git pull
+	cd .. && git clone https://github.com/percona/proxysql_exporter || cd ./proxysql_exporter && git pull
+	ls -la .. | grep _exporter
```
BupycHuk, line 38: probably it can be replaced by downloading latest version of exporter binaries instead of cloning repositories

YaroslavPodorvanov: adding this part for faster development for beginner in PMM project like me

ritbl: @YaroslavPodorvanov  I agree with @BupycHuk, it is up to a developer what way they want to pull exporters, via `source code` or binaries. If you clone code you might need extra configuration to build/install it.

ritbl: Additionally, you need to install it, just clone it.
So that it is available in `$PATH` (via `$GOPATH/bin`)

ritbl (changes requested):
please see my comments

YaroslavPodorvanov:
Will re-open after full understand project need

PR #342, opened 2022-03-23
Pmm 5680 store logs

qwest812:
https://jira.percona.com/browse/PMM-5680

### DIFF
--- a/agents/supervisor/supervisor.go
+++ b/agents/supervisor/supervisor.go
@@ -19,6 +19,7 @@ package supervisor
 import (
 	"context"
 	"fmt"
+	"github.com/percona/pmm-agent/storelogs"
 	"path/filepath"
 	"regexp"
 	"runtime/pprof"
@@ -345,8 +346,9 @@ func (s *Supervisor) startProcess(agentID string, agentProcess *agentpb.SetState
 		"type":      agentType,
 	})
 	l.Debugf("Starting: %s.", processParams)
-
-	process := process.New(processParams, agentProcess.RedactWords, l)
+	ringLog := new(storelogs.LogsStore)
+	ringLog.SetUp(100, l)
+	process := process.New(processParams, agentProcess.RedactWords, l, ringLog)
 	go pprof.Do(ctx, pprof.Labels("agentID", agentID, "type", agentType), process.Run)
 
 	done := make(chan struct{})
--- a/storelogs/main.go
+++ b/storelogs/main.go
@@ -3,36 +3,43 @@ package storelogs
 import (
 	"container/ring"
 	"fmt"
+	"github.com/sirupsen/logrus"
 )
 
 type LogsStore struct {
-	MapLogs map[string]*ring.Ring
+	log   *ring.Ring
+	entry *logrus.Entry
 }
 
-func (l *LogsStore) SetUp(exporter string, countLogs int) {
-	l.MapLogs[exporter] = ring.New(countLogs)
+func (l *LogsStore) SetUp(countLogs int, entry *logrus.Entry) {
+	l.log = ring.New(countLogs)
+	l.entry = entry
 }
 
-func (l *LogsStore) LenLogs(exporter string) int {
-	r, ok := l.MapLogs[exporter]
-	if !ok {
-		return 0
+//func (l *LogsStore) LenLogs(exporter string) int {
+//	r, ok := l.MapLogs[exporter]
+//	if !ok {
+//		return 0
+//	}
+//	return r.Len()
+//}
+
+func (l *LogsStore) SaveLog(log string) {
+	log = l.entry.Level.String() + ": " + log
+	for _, v := range l.entry.Data {
+		log = log + fmt.Sprintf("       %v", v)
 	}
-	return r.Len()
+	l.log.Value = log
+	l.log = l.log.Next()
+	//l.MapLogs[exporter].Value = log
+	//l.MapLogs[exporter] = l.MapLogs[exporter].Next()
 }
 
-func (l *LogsStore) SaveLog(exporter string, log string) {
-	l.MapLogs[exporter].Value = log
-	l.MapLogs[exporter] = l.MapLogs[exporter].Next()
-}
-
-func (l *LogsStore) GetLogs(exporter string) []string {
-	var logs []string
-	l.MapLogs[exporter].Do(func(p interface{}) {
-		log := fmt.Sprint(p)
-		logs = append(logs, log)
-	})
-	return logs
-}
-
-var Logs = new(LogsStore)
+//func (l *LogsStore) GetLogs(exporter string) []string {
+//	var logs []string
+//	l.MapLogs[exporter].Do(func(p interface{}) {
+//		log := fmt.Sprint(p)
+//		logs = append(logs, log)
+//	})
+//	return logs
+//}
--- a/storelogs/main.go
+++ b/storelogs/main.go
@@ -1,0 +1,38 @@
+package storelogs
+
+import (
+	"container/ring"
+	"fmt"
+)
+
+type LogsStore struct {
+	MapLogs map[string]*ring.Ring
+}
+
+func (l *LogsStore) SetUp(exporter string, countLogs int) {
+	l.MapLogs[exporter] = ring.New(countLogs)
+}
+
+func (l *LogsStore) LenLogs(exporter string) int {
+	r, ok := l.MapLogs[exporter]
+	if !ok {
+		return 0
+	}
+	return r.Len()
+}
+
+func (l *LogsStore) SaveLog(exporter string, log string) {
+	l.MapLogs[exporter].Value = log
+	l.MapLogs[exporter] = l.MapLogs[exporter].Next()
+}
+
+func (l *LogsStore) GetLogs(exporter string) []string {
+	var logs []string
+	l.MapLogs[exporter].Do(func(p interface{}) {
+		log := fmt.Sprint(p)
+		logs = append(logs, log)
+	})
+	return logs
+}
+
+var Logs = new(LogsStore)
--- a/agents/process/process.go
+++ b/agents/process/process.go
@@ -17,9 +17,9 @@
 package process
 
 import (
-	"container/ring"
 	"context"
 	"fmt"
+	"github.com/percona/pmm-agent/storelogs"
 	"os/exec"
 	"strings"
 	"time"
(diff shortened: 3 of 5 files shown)


storelogs/main.go
```diff
@@ -0,0 +1,39 @@
...
+func (l *LogsStore) SetUp(countLogs int, entry *logrus.Entry) {
+	l.log = ring.New(countLogs)
+	l.entry = entry
+}
+
+func (l *LogsStore) SaveLog(log string) {
+	dt := time.Now()
+	log = dt.Format("01-02-2006 15:04:05") + " " + l.entry.Level.String() + ": " + log
+	for _, v := range l.entry.Data {
+		log = log + fmt.Sprintf("    %v", v)
```
YaroslavPodorvanov, line 24: 1 solution make slice of strings and then join
2 solution by any buffer
in `log = log + fmt.Sprintf(" %v", v)` too many garbage https://gosamples.dev/concatenate-strings/

agentlocal/agent_local.go
```diff
@@ -63,6 +63,11 @@ type Server struct {
 	reloadCloseOnce sync.Once
 }
 
+func (s *Server) mustEmbedUnimplementedAgentLocalServer() {
```
YaroslavPodorvanov, line 66: it's wrong, possible solution remove this requirement by some protoc --arg or by embeding [agentlocalpb|agentpb].UnimplementedAgentLocalServer

storelogs/main.go
qwest812: changed to buffer

storelogs/main.go
```diff
@@ -0,0 +1,84 @@
...
+	"bytes"
+	"container/ring"
+	"fmt"
+	"github.com/sirupsen/logrus"
+	"time"
+)
+
+type LogsStore struct {
+	log   *ring.Ring
+	Entry *logrus.Entry
```
YaroslavPodorvanov, line 13: entry *logrus.Entry

YaroslavPodorvanov (changes requested):

storelogs/main.go
qwest812: I need use in here newProcessLogger(sl.Entry, keepLogLines, redactWords),

YaroslavPodorvanov (changes requested):
mustEmbed need delete

YaroslavPodorvanov (approved):

storelogs/main.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// LogsStore implement ring save logs
+type LogsStore struct {
+	log   *ring.Ring
```
BupycHuk, line 31: we already have ring logic on our process side, so we can just request it from there.

storelogs/main.go
```diff
@@ -0,0 +1,117 @@
...
+	l.log = ring.New(l.count)
+	l.Entry = entry
+}
+
+// SetCountLogs sets the number of logs to store
+func (l *LogsStore) SetCountLogs(countLogs int) {
+	l.count = countLogs
+}
+
+func (l *LogsStore) saveLog(log string) {
```
BupycHuk, line 51: probably we could do it using Hooks mechanism or by replacing output in logrus, so we won't be need to do that.

storelogs/main.go
```diff
@@ -0,0 +1,107 @@
...
+	l.Entry = entry
+}
+
+func (l *LogsStore) SetCountLogs(countLogs int) {
+	l.count = countLogs
+}
+
+func (l *LogsStore) SaveLog(log string) {
+	dt := time.Now()
+	var b bytes.Buffer
```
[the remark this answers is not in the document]
BupycHuk: minor: let's disable varnamelen or tweak it.

storelogs/main.go
qwest812: > we already have ring logic on our process side, so we can just request it from there.

where, I can't find it

storelogs/main.go
qwest812: > probably we could do it using Hooks mechanism or by replacing output in logrus, so we won't be need to do that.


this is not an exported function and they work with standard logrus functions, I didn’t manage to add this when outputting, so I made wrapper functions like Infof, Debugf etc..

agents/supervisor/supervisor.go
```diff
@@ -119,11 +122,13 @@ func (s *Supervisor) AgentsList() []*agentlocalpb.AgentInfo {
 	res := make([]*agentlocalpb.AgentInfo, 0, len(s.agentProcesses)+len(s.builtinAgents))
 
 	for id, agent := range s.agentProcesses {
+		logs := agent.logs.GetLogs()
```
BupycHuk, line 125: in scope of this task we don't have requirement to split logs per agent, but ok.
QAN(bultin) agents logs are missed

storelogs/main.go
```diff
@@ -0,0 +1,117 @@
...
+// LogsStore implement ring save logs.
+type LogsStore struct {
+	log   *ring.Ring
+	Entry *logrus.Entry
+	count int
+	m     sync.RWMutex
+}
+
+// Init initializes basic parameters.
+func (l *LogsStore) Init(entry *logrus.Entry) {
```
BupycHuk, line 38: we use another way to create and initialize objects through functions `New` or `NewSomething` which returns initialized object.

storelogs/main.go
```diff
@@ -0,0 +1,117 @@
...
+		l.count = 10
+	}
+	l.log = ring.New(l.count)
+	l.Entry = entry
+}
+
+// SetCountLogs sets the number of logs to store.
+func (l *LogsStore) SetCountLogs(countLogs int) {
+	l.count = countLogs
+}
```
BupycHuk, line 49: this method doesn't makes sense, because it doesn't change capacity of ring

storelogs/main.go
BupycHuk: I prepared PR with changes I suggested https://github.com/percona/pmm-agent/pull/364, Please take a look

agents/supervisor/supervisor.go
BupycHuk: I'm surprised why it was made this way,  although there was suggested implementation with a separate endpoint for logs. This implementation doesn't include all pmm-agent related logs(like logs related to connection to pmm-managed).

BupycHuk (changes requested):

storelogs/main.go
qwest812: you are right

commands/run.go
```diff
@@ -77,7 +78,8 @@ func run(ctx context.Context, cfg *config.Config, configFilepath string) {
 	connectionChecker := connectionchecker.New(&cfg.Paths)
 	v := versioner.New(&versioner.RealExecFunctions{})
 	client := client.New(cfg, supervisor, connectionChecker, v)
-	localServer := agentlocal.NewServer(cfg, supervisor, client, configFilepath)
+	ringLog := storelogs.New(500)
```
BupycHuk, line 81: let's move it outside the `run` function because `run` is called each time we lose connection with pmm-server

agentlocal/agent_local.go
```diff
@@ -289,6 +298,42 @@ func (s *Server) runJSONServer(ctx context.Context, grpcAddress string) {
 	mux.Handle("/debug/", http.DefaultServeMux)
 	mux.Handle("/debug", debugPageHandler)
 	mux.Handle("/", proxyMux)
+	mux.HandleFunc("/logs.zip", func(w http.ResponseWriter, r *http.Request) {
```
BupycHuk, line 301: can we add a test for logs.zip?

agentlocal/agent_local_test.go
```diff
@@ -108,3 +116,91 @@ func TestServerStatus(t *testing.T) {
...
+		_, err := s.Status(context.Background(), &agentlocalpb.StatusRequest{GetNetworkInfo: false})
+		require.NoError(t, err)
+		// fmt.Sprintf("%v %v", actual, err)
+		rec := httptest.NewRecorder()
+		req := httptest.NewRequest("GET", "/logs.zip", nil)
+		s.Zip(rec, req)
+		b, err := ioutil.ReadAll(rec.Body)
+		require.NoError(t, err)
+
+		expectedFile, err := generateTestZip(s)
```
ritbl, line 168: You are generating zip file that you will later compare with, but what if it fails? The output will be not readable.
I think it would be better to unzip received file, and verify content.
Here is example of equality check failing:
image

The failure message is impossible to understand, because it is a diff between two binary files

ritbl (changes requested):
please see my comment

agentlocal/agent_local.go
```diff
@@ -61,23 +65,28 @@ type Server struct {
...
+	ringLogs        *storelogs.LogsStore
 	reload          chan struct{}
 	reloadCloseOnce sync.Once
 
 	agentlocalpb.UnimplementedAgentLocalServer
 }
 
 // NewServer creates new server.
-//
+
```
BupycHuk, line 76: let's revert it.

agentlocal/agent_local.go
```diff
@@ -314,7 +324,53 @@ func (s *Server) runJSONServer(ctx context.Context, grpcAddress string) {
...
+	buf := &bytes.Buffer{}
+	writer := zip.NewWriter(buf)
+	b := &bytes.Buffer{}
+	for _, serverLog := range s.ringLogs.GetLogs() {
+		_, err := b.WriteString(serverLog)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+	addData(writer, "server.txt", b.Bytes())
```
BupycHuk, line 354: I think it's better to rename it to `pmm-agent.txt` or `http-server` if it contains only http server logs.
```suggestion
	addData(writer, "http-server.txt", b.Bytes())
```

storelogs/storelogs.go
```diff
@@ -0,0 +1,63 @@
...
+		log: ring.New(count),
+		m:   sync.RWMutex{},
+	}
+}
+
+// Write writes log for store.
+func (l *LogsStore) Write(b []byte) (n int, err error) {
+	l.m.Lock()
+	l.log.Value = string(b)
+	l.m.Unlock()
```
BupycHuk, line 44: let's unlock it one line below.

BupycHuk (approved):

PR #347, opened 2022-03-25
PMM-9415 fix first restore attempt never works

PavelKhripkov:
PMM-9415

Build: SUBMODULES-2402

- [ ] https://github.com/percona/pmm-managed/pull/1059


jobs/mysql_restore_job.go
```diff
@@ -35,9 +35,10 @@ import (
 
 const (
 	xbstreamBin      = "xbstream"
-	mySQLServiceName = "mysql"
+	mySQLServiceName = "mysqld"
```
artemgavrilov, line 38: I wonder how it worked before? 🤔
Why we faced this bug only during first restore?

jobs/mysql_restore_job.go
```diff
@@ -341,7 +342,17 @@ func isPathExists(path string) (bool, error) {
...
+func getPermissions(path string) (os.FileMode, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return 0, err
+	}
+	return info.Mode(), nil
+}
+
 func restoreBackup(ctx context.Context, backupDirectory, mySQLDirectory string) error {
+	mysqlDirPermissions := os.FileMode(0751)
```
artemgavrilov, line 354: Where did you get this permissions? If from some MySQL docs then let's add link to the comments.

jobs/mysql_restore_job.go
```diff
@@ -386,5 +401,9 @@ func restoreBackup(ctx context.Context, backupDirectory, mySQLDirectory string)
 		return errors.WithStack(err)
 	}
 
+	if err := os.Chmod(mySQLDirectory, mysqlDirPermissions); err != nil {
+		return errors.Wrap(err, "failed to change permissions for MySQL base directory")
+	}
```
artemgavrilov, line 406: I guess it should be recursive as well. Same as we do few lines above with chown, no?

jobs/mysql_restore_job.go
PavelKhripkov: Briefly, due to we don't stop the service (because we check the wrong one) file /var/lib/mysql/mysql.sock is not recreated in correct place. So, the restoring being processed correctly, we just cannot connect the DB. On the second try it works other way because we stop the service (stopping mysql causes stopping mysqld).

jobs/mysql_restore_job.go
PavelKhripkov: I copied it from default mysql folder. It is used only if this folder is absent (sth gone wrong with DB). Such a way we provide an access to DB and all the permissions should be rechecked by DBA who is responsible for this DB. By default, during the restore, permissions are set to 0750 which may cause users cannot connect.

jobs/mysql_restore_job.go
PavelKhripkov: To allow all users to connect to the DB, we only need to change the permissions of the parent directory. All permissions inside the main directory are controlled by xtrabackup. So I don't think we need to change it recursively as it works without any problems.

idoqo (approved):

BupycHuk (approved):

jobs/mysql_restore_job.go
BupycHuk: I think it's good to add comment into the code why we set these permissions

PavelKhripkov: Done.

jobs/mysql_restore_job.go
artemgavrilov: Got it, thanks!

artemgavrilov:
@PavelKhripkov Danm, look what I found: https://dev.mysql.com/doc/refman/8.0/en/using-systemd.html#systemd-overview

So RPM based systems have `mysqld` service name, and Debian based systems use `mysql`.

What distro did you use for testing?

PavelKhripkov:
@artemgavrilov really damn) I used Amazon Linux as described in the tkt.
Thanks! I'll test it on different platforms and make changes if needed.

jobs/mysql_restore_job.go
PavelKhripkov: After initial installation, main mysql directory has permissions 0751, and so I can use shell `mysql` command to connect to the DB under non-root user. After restoring, if we don't change permissions, `xtrabackup` utility sets permissions to 0750, therefore the DB becomes inaccessible for non-root users.

As per documentation you provided there are no recommendations for the main folder. So we can set +x to allow to see the  content of the dir. What do you think?

PavelKhripkov: I didn't get an exact answer, so I changed permissions to 0750 and left a TODO in order we can address this problem when we start close work on the backup service.

BupycHuk (approved):

idoqo (approved):

jobs/mysql_restore_job.go
```diff
@@ -34,14 +34,18 @@ import (
...
+	mySQLSystemUserName  = "mysql"
+	mySQLSystemGroupName = "mysql"
+	// TODO make mySQLDirectory autorecognized as done in 'xtrabackup' utility; see 'xtrabackup --help' --datadir parameter
 	mySQLDirectory   = "/var/lib/mysql"
 	systemctlTimeout = 10 * time.Second
 )
 
+var (
+	mysqlServiceRegex = regexp.MustCompile(`mysql(d)?\.service`) // this is used to lookup MySQL service in the list of all system services
+)
```
artemgavrilov, line 47:
```suggestion
var mysqlServiceRegex = regexp.MustCompile(`mysql(d)?\.service`) // this is used to lookup MySQL service in the list of all system services
```

artemgavrilov (approved):
Nice!

PR #372, opened 2022-05-03
PMM-9320: fix username/password incorrect escape

ritbl:
PMM-9320

Password and Username are escaped According to RFC 3986 §3.2.1

// The RFC allows ';', ':', '&', '=', '+', '$', and ',' in
// userinfo, so we must escape only '@', '/', and '?'.
// The parsing of userinfo treats ':' as special so we must escape
// that too.
go1.18.1/src/net/url/url.go:143

But in MongoDB driver they decided to unscape it as encodePathSegment
using url.PathUnescape
mongo-go-driver/x/mongo/driver/connstring/connstring.go:236 
and
ext/mongo-go-driver/x/mongo/driver/connstring/connstring.go:249

### DIFF
--- a/agents/mongodb/internal/profiler/collector/collector_test.go
+++ b/agents/mongodb/internal/profiler/collector/collector_test.go
@@ -18,6 +18,7 @@ package collector
 import (
 	"context"
 	"fmt"
+	"github.com/percona/pmm-agent/mongo_fix"
 	"strings"
 	"sync"
 	"testing"
@@ -30,7 +31,6 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
-	"go.mongodb.org/mongo-driver/mongo/options"
 	"go.mongodb.org/mongo-driver/mongo/readpref"
 )
 
@@ -202,8 +202,12 @@ func createSession(dsn string, agentID string) (*mongo.Client, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), MgoTimeoutDialInfo)
 	defer cancel()
 
-	opts := options.Client().
-		ApplyURI(dsn).
+	opts, err := mongo_fix.ClientForDSN(dsn)
+	if err != nil {
+		panic(err)
+	}
+
+	opts = opts.
 		SetDirect(true).
 		SetReadPreference(readpref.Nearest()).
 		SetSocketTimeout(MgoTimeoutSessionSocket).
--- a/mongo_fix/mongo_fix.go
+++ b/mongo_fix/mongo_fix.go
@@ -1,0 +1,25 @@
+package mongo_fix
+
+import (
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"net/url"
+)
+
+// ClientForDSN applies URI to Client
+func ClientForDSN(dsn string) (*options.ClientOptions, error) {
+	parsedDsn, err := url.Parse(dsn)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot parse DSN")
+	}
+
+	username := parsedDsn.User.Username()
+	password, _ := parsedDsn.User.Password()
+	creds := options.Credential{Username: username, Password: password}
+
+	clientOptions := options.Client().
+		ApplyURI(dsn).
+		SetAuth(creds) //Workaround for PMM-9320
+
+	return clientOptions, nil
+}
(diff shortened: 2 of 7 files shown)


mongo_fix/mongo_fix.go
```diff
@@ -0,0 +1,25 @@
...
+		return nil, errors.Wrap(err, "cannot parse DSN")
+	}
+
+	username := parsedDsn.User.Username()
+	password, _ := parsedDsn.User.Password()
+	creds := options.Credential{Username: username, Password: password}
+
+	clientOptions := options.Client().
+		ApplyURI(dsn).
+		SetAuth(creds) //Workaround for PMM-9320
```
ritbl, line 22: fixing here

agents/mongodb/internal/profiler/collector/collector_test.go
```diff
@@ -202,8 +203,12 @@ func createSession(dsn string, agentID string) (*mongo.Client, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), MgoTimeoutDialInfo)
 	defer cancel()
 
-	opts := options.Client().
-		ApplyURI(dsn).
+	opts, err := mongo_fix.ClientForDSN(dsn)
+	if err != nil {
+		panic(err)
```
ritbl, line 208: it is test, so we can `panic`

mongo_fix/mongo_fix.go
```diff
@@ -0,0 +1,27 @@
+package mongo_fix
```
ritbl, line 1: called `mongo_fix` not to clash with mongo driver

mongo_fix/mongo_fix.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot parse DSN")
+	}
+
+	clientOptions := options.Client().ApplyURI(dsn)
+
+	// Workaround for PMM-9320
+	username := parsedDsn.User.Username()
+	password, _ := parsedDsn.User.Password()
+	if username != "" || password != "" {
```
ritbl, line 22: we need to set `auth` only when creds are provided, otherwise auth fill fail

utils/tests/mongodb.go
```diff
@@ -104,7 +105,12 @@ func GetTestMongoDBReplicatedWithSSLDSN(tb testing.TB, pathToRoot string) (strin
 func OpenTestMongoDB(tb testing.TB, dsn string) *mongo.Client {
 	tb.Helper()
 
-	client, err := mongo.Connect(context.Background(), options.Client().ApplyURI(dsn))
+	opts, err := mongo_fix.ClientForDSN(dsn)
+	if err != nil {
+		panic(err)
```
ritbl, line 110: `panic` in test

mongo_fix/mongo_fix.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"net/url"
+
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+// ClientForDSN applies URI to Client.
+func ClientForDSN(dsn string) (*options.ClientOptions, error) {
```
BupycHuk, line 26: probably we should call it `ClientOptionsForDSN` or include `mongo.Connect` to this function

BupycHuk (commented):
Looks good to me. Can we add a few tests with this kind of DSNs?

agents/mongodb/internal/profiler/collector/collector_test.go
ritbl: okay, here we return error, so I think we can just propagate error.
but what about other cases, like in `func OpenTestMongoDB(tb testing.TB, dsn string) *mongo.Client {`
we don't have error in the API, do you think we should refactor API or we can panic (in tests only)

mongo_fix/mongo_fix.go
ritbl: yeah, `ClientOptions` is more correct

ritbl:
> Looks good to me. Can we add a few tests with this kind of DSNs?
sure, will do

agents/mongodb/internal/profiler/collector/collector_test.go
BupycHuk: we can call `require.NoError(tb, err)` in that places. so it will stop test with failed result

ritbl: yeah, good idea

BupycHuk (approved):

artemgavrilov (approved):

PR #374, opened 2022-05-04
PMM-9632 Settings view usage.

JiriCtvrtka:
PMM-9632

Build: SUBMODULES-0

- [ ] Links to other linked pull requests (optional).

### DIFF
--- a/agents/supervisor/supervisor.go
+++ b/agents/supervisor/supervisor.go
@@ -66,10 +66,11 @@ type Supervisor struct {
 
 // agentProcessInfo describes Agent process.
 type agentProcessInfo struct {
-	cancel         func()          // to cancel Process.Run(ctx)
-	done           <-chan struct{} // closes when Process.Changes() channel closes
-	requestedState *agentpb.SetStateRequest_AgentProcess
-	listenPort     uint16
+	cancel          func()          // to cancel Process.Run(ctx)
+	done            <-chan struct{} // closes when Process.Changes() channel closes
+	requestedState  *agentpb.SetStateRequest_AgentProcess
+	listenPort      uint16
+	processExecPath string
 }
 
 // builtinAgentInfo describes built-in Agent.
@@ -120,10 +121,11 @@ func (s *Supervisor) AgentsList() []*agentlocalpb.AgentInfo {
 
 	for id, agent := range s.agentProcesses {
 		info := &agentlocalpb.AgentInfo{
-			AgentId:    id,
-			AgentType:  agent.requestedState.Type,
-			Status:     s.lastStatuses[id],
-			ListenPort: uint32(agent.listenPort),
+			AgentId:         id,
+			AgentType:       agent.requestedState.Type,
+			Status:          s.lastStatuses[id],
+			ListenPort:      uint32(agent.listenPort),
+			ProcessExecPath: agent.processExecPath,
 		}
 		res = append(res, info)
 	}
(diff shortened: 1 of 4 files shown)


agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -92,7 +81,7 @@ const (
 	commandTypeUpdate       = "UPDATE"
 	commandTypeInsert       = "INSERT"
 	commandTypeDelete       = "DELETE"
-	commandTypeUtiity       = "UTILITY"
+	commandTypeUtility      = "UTILITY"
```
JiriCtvrtka, line 84: Already fixed in previous commit, but its not merged yet.

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -41,24 +40,14 @@ import (
...
 // PGStatMonitorQAN QAN services connects to PostgreSQL and extracts stats.
 type PGStatMonitorQAN struct {
-	q            *reform.Querier
-	dbCloser     io.Closer
-	agentID      string
-	l            *logrus.Entry
-	changes      chan agents.Change
-	monitorCache *statMonitorCache
-
-	// By default, query shows the actual parameter instead of the placeholder.
```
JiriCtvrtka, line 55: Since pgsmNormalizedQuery and waitTime can be changed without restart its not good idea to cache it anymore. Now we checking settings everytime when we creating new bucket.

agents/postgres/pgstatmonitor/models.go
```diff
@@ -49,8 +49,14 @@ type pgStatMonitorSettings struct {
...
-	Name  string `reform:"name"`
-	Value string `reform:"value"`
+	Name         string  `reform:"name"`
+	Value        string  `reform:"value"`
+	DefaultValue string  `reform:"default_value"`
+	Description  string  `reform:"description"`
+	Minimum      *int64  `reform:"minimum"`
+	Maximum      *int64  `reform:"maximum"`
+	Options      *string `reform:"options"`
+	Restart      string  `reform:"restart"`
```
BupycHuk, line 59: let's keep only fields that we use in case PGSM team changes structure later we will have more PMM clients supported latest PGSM

agents/postgres/pgstatmonitor/pgstatmonitor.go
BupycHuk: 👍🏼

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -307,12 +242,12 @@ func (m *PGStatMonitorQAN) Run(ctx context.Context) {
 				m.changes <- agents.Change{Status: inventorypb.AgentStatus_STARTING}
 			}
 
-			lengthS := uint32(m.waitTime.Seconds())
+			lengthS := uint32(waitTime.Seconds())
 			buckets, err := m.getNewBuckets(ctx, lengthS)
```
BupycHuk, line 246: I think you can call settings here and pass them to the `getNewBuckets` method, in that case, you will be able to support immediate change for `waitTime` as well.

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -331,8 +266,80 @@ func (m *PGStatMonitorQAN) Run(ctx context.Context) {
...
+	}
+
+	return time.Duration(valueInt) * time.Second, nil
+}
+
 func (m *PGStatMonitorQAN) getNewBuckets(ctx context.Context, periodLengthSecs uint32) ([]*agentpb.MetricsBucket, error) {
-	current, prev, err := m.monitorCache.getStatMonitorExtended(ctx, m.q, m.pgsmNormalizedQuery)
+	settings, err := m.getSettings()
+	if err != nil {
+		m.l.Errorf(err.Error())
```
BupycHuk, line 335: If we get an error here, shouldn't we stop the method and return an error?
The same below

JiriCtvrtka: Good catch. Moved before line 245 in pgstatmonitor.go as you required above and added continue there.

agents/postgres/pgstatmonitor/pgstatmonitor.go
JiriCtvrtka: Done.

agents/postgres/pgstatmonitor/models.go
JiriCtvrtka: Done.

agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -331,8 +281,71 @@ func (m *PGStatMonitorQAN) Run(ctx context.Context) {
...
+	return settings, nil
+}
+
+func (s settings) getNormalizedQueryValue() (bool, error) {
+	key := "pg_stat_monitor.pgsm_normalized_query"
+	if _, ok := s[key]; !ok {
+		return false, errors.New("failed to get pgsm_normalized_query property")
+	}
+
+	if s[key].Value == "yes" {
```
BupycHuk, line 326: If I got it right this value might be "1" as well

JiriCtvrtka: Yes string "1" for older versions. Thank you.

BupycHuk (approved):

ShashankSinha252 (approved):

percona-csalguero (approved):

PR #380, opened 2022-05-10
PMM-9921 rewrite pmm client image entrypoint

nikita-b:
PMM-9921

It's rewriting entrypoint.py for pmm-client contianer in Go.
Here original:  https://github.com/Percona-Lab/pmm-submodules/blob/PMM-2.0/build/pmm-client-docker/entrypoint.py

Build: SUBMODULES-0

- [ ] Links to other linked pull requests (optional).


cmd/pmm-agent-entrypoint/main.go
```diff
@@ -0,0 +1,279 @@
...
+		preSIGKILLtimeout := 10
+		timer := sendSIGKILLwithTimeout(agent.Process, preSIGKILLtimeout, l)
+
+		err = agent.Wait()
+		if err != nil {
+			exitError, ok := err.(*exec.ExitError)
+			if !ok {
+				l.Warnf("Can't get exit code for pmm-agent. Error code: %s", err)
+			}
+			l.Infof("Prerun pmm-agent exited with %d", exitError.ExitCode())
```
BupycHuk, line 266: in case of `!ok` it will `panic`, so it's better to put it to `else`

cmd/pmm-agent-entrypoint/main.go
```diff
@@ -0,0 +1,279 @@
...
+var pmmAgentProcessID = 0
+
+func getEnvWithDefault(key, defautlValue string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return defautlValue
+}
+
+func runPmmAgent(commandLineArgs []string, restartPolicy restartPolicy, l *logrus.Entry, pmmAgentSidecarSleep int) int {
```
BupycHuk, line 75: let's pass context here, so we can be sure that we stopped this `for` loop.

BupycHuk: there we can use something like
```
for {
    select {
        case <- ctx.Done():
        default: // or might be some timer channel
    }
}
```

BupycHuk (changes requested):

cmd/pmm-agent-entrypoint/main.go
```diff
@@ -0,0 +1,271 @@
...
+		s := <-signals
+		signal.Stop(signals)
+		l.Warnf("Got %s, shutting down...", unix.SignalName(s.(unix.Signal)))
+		if pmmAgentProcessID != 0 {
+			// gracefull shutdown for pmm-agent
+			if err := syscall.Kill(pmmAgentProcessID, syscall.SIGTERM); err != nil {
+				l.Warn("Failed to send SIGTERM, command must have exited:", err)
+			}
+		}
+		cancel()
```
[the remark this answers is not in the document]
nikita-b: We can't use `CommandContext` because it sends SIGKILL to subprocess. I used it for some command but not for `pmm-agent`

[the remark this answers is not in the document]
BupycHuk: and here we exactly send SIGKILL command.
https://github.com/percona/pmm-agent/pull/380/files#diff-362c1b9d32b751c3986e334de58ca698848712a7fa4cf893424e9deac20d59a3R117-R123
So I suggested to use context instead of trying to extract PID and sending signal manually

BupycHuk (approved):

cmd/pmm-agent-entrypoint/main.go
```diff
@@ -0,0 +1,267 @@
...
+	"os/exec"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	reaper "github.com/ramr/go-reaper"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+	"gopkg.in/alecthomas/kingpin.v2"
```
atymchuk, line 30: @BupycHuk Can we start using cobra here? The comparison doesn't look promising for kingpin https://go.libhunt.com/compare-kingpin-vs-cobra

atymchuk (changes requested):

cmd/pmm-agent-entrypoint/main.go
nikita-b: We use kingpin everywhere in PMM. If we want to use cobra we need to migrate pmm-agent entirely. We want to migrate to a monorepo, it seems not the best idea to add new dependencies now.

atymchuk: How can you migrate then a component that has not been maintained for 4 years? What strategy do you suggest? Never is not a strategy I'm afraid.

BupycHuk: Not as part of this PR, we started using kong in mongodb_exporter and if we decide to migrate to another lib we should make a decision what tool will be used and will it be backward compatible with the kingpin.

What do you mean by saying how can we migrate component that hasn't been maintained? We will migrate pmm-agent which has kingpin as dependency and based on current state of project it does everything required from it, so it's not a blocker for migration.

atymchuk: Got it, thanks!

atymchuk (approved):

ademidoff:
It was ported to the monorepo.

